PR #8, opened 2023-12-13
[TC-066] gh workflows for PR

chelsea-EYDS:
I will follow up with another PR for the frontend test containers - this will install the backend modules, verify the container can build, and runs the tests (tests TODO - this just verifies that the we can build/run)


chelsea-EYDS:
Changes are on another branch

PR #10, opened 2023-12-16
[TC-0059] Backend Auth + Roles Check

chelsea-EYDS:
- Checking auth token on all routes that are not annotated with the public decorator
- Validating auth token
- Checking Reflector to see if route is annotated with a Roles decorator
- Validating user assigned role if the current route requires a role


chelsea-EYDS:
changing branches in order to link to jira

PR #11, opened 2023-12-16
[TC-0067] (pt1) added cypress to frontend

chelsea-EYDS:
- cypress added to frontend
- (pipeline config to follow)


chelsea-EYDS:
Changing branch name to link to jira

PR #12, opened 2023-12-16
[TC-0066] (pt1) jest config

chelsea-EYDS:
- Added backend config for jest
- Runs on local, pipeline config to follow


chelsea-EYDS:
Changing branch name in order to link to jira

PR #13, opened 2023-12-16
[TC-0066] (pt2) pipeline config for running tests

chelsea-EYDS:
- dynamically setting the ENV variable when running `make build-test` or `make build-local`
- If ENV=ci the frontend builds with the cypress base image


chelsea-EYDS:
Changing branch name to link to Jira

PR #14, opened 2023-12-16
[TC-0066] (pt3) pipeline config

chelsea-EYDS:


fw-noel (approved):
anything we can do about the warnings on GHA? `lint-frontend`, etc.

Oh jeez thanks for catching that. Yes, all resolved.

PR #16, opened 2023-12-16
[TC-0059] Auth Guard

chelsea-EYDS:
- Checking auth token on all routes that are not annotated with the public decorator
- Validating auth token
- Checking Reflector to see if route is annotated with a Roles decorator
- Validating user assigned role if the current route requires a role

### DIFF
--- a/backend/src/auth/auth.guard.ts
+++ b/backend/src/auth/auth.guard.ts
@@ -1,54 +1,47 @@
 import {
+  Injectable,
   CanActivate,
   ExecutionContext,
-  ForbiddenException,
-  Injectable,
   UnauthorizedException,
 } from '@nestjs/common';
 import { Reflector } from '@nestjs/core';
-import { Request } from 'express';
-import jwt from 'jsonwebtoken';
-import { REALM, AUTH_SERVER } from './const';
-import { Token } from './interface';
-import { IS_PUBLIC_KEY } from './public.decorator';
+import jwt, { JwtPayload } from 'jsonwebtoken';
+
+import { AUTH_CLIENT, AUTH_SERVER, AUTH_REALM } from './const';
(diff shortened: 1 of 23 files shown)


backend/src/auth/auth.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    if (payload.iss !== `${AUTH_SERVER}/realms/${AUTH_REALM}`) {
+      throw new UnauthorizedException();
+    }
+
+    if (payload.aud !== 'account') {
+      throw new UnauthorizedException();
+    }
+
+    //YODO check expiry
```
fw-noel, line 76: This would be pretty trivial, no?

backend/src/auth/auth.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (!token) {
+      throw new UnauthorizedException();
+    }
+    const payload = token as jwt.JwtPayload;
+
+    if (payload.iss !== `${AUTH_SERVER}/realms/${AUTH_REALM}`) {
+      throw new UnauthorizedException();
+    }
+
+    if (payload.aud !== 'account') {
```
fw-noel, line 72: This seems like the wrong thing to check? I think authorized party is the thing to check `azp`

backend/src/auth/public.decorator.ts
```diff
@@ -0,0 +1,4 @@
+import { SetMetadata } from '@nestjs/common';
+import { Metadata } from './metadata';
+
+export const Public = () => SetMetadata(Metadata.IS_PUBLIC, true);
```
fw-noel, line 4: can we change to `PUBLIC_ENDPOINT` just for clarity

fw-noel (approved):

PR #21, opened 2023-12-19
[TC-58] Spec out the data model

fw-noel:
[TC-58](https://bcdevex.atlassian.net/browse/TC-58)

Objective:
We're adding the entities and enums that we should need for the initial step of the data model. This can be further refined, and later created as a migration.

Currently, I'm leaving the entities in the database folder until we start building out controllers and whatnot. It can be moved from there.

[TC-58]: https://bcdevex.atlassian.net/browse/TC-58?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/src/common/enums/availability-type.enum.ts
```diff
@@ -0,0 +1,4 @@
+export enum AvailabilityType {
+  AVAILABILITY,
```
chelsea-EYDS, line 2: AVAILABLE?

I think we 'UNAVAILABLE' as well? (ie: not deployed but they're on holiday or something)

backend/src/common/enums/experience.enum.ts
```diff
@@ -0,0 +1,6 @@
+export enum Experience {
+  INTERESTED,
```
chelsea-EYDS, line 2: Do we need to set the string value here? (I'm sure this is right but this part of typescript confuses me still)

chelsea-EYDS: (I just wasn't sure if they would be initialized w/o the  = "enum string value")

backend/src/database/entities/availability.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+  PrimaryGeneratedColumn,
+  ManyToOne,
+  JoinColumn,
+} from 'typeorm';
+import { AvailabilityType } from '../../common/enums/availability-type.enum';
+import { PersonnelEntity } from './personnel.entity';
+
+@Entity('function')
+export class AvailabilityEntity {
+  @PrimaryGeneratedColumn('increment')
```
chelsea-EYDS, line 13: number ids instead of uuid? (I'm fine with it just checking if this is the result of setting 'increment')

backend/src/database/entities/training.entity.ts
```diff
@@ -0,0 +1,14 @@
...
+  Entity,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+
+@Entity('training')
+export class TrainingEntity {
+  @PrimaryGeneratedColumn('increment')
+  id: string;
+
+  @Column({ name: 'name', type: 'varchar', length: 100 })
```
chelsea-EYDS, line 12: Will we add a new 'training' type every time a user is added (or their training updated in future)? 

Or will this be a static table with predefined trainings that we can map to the resource?

backend/src/database/entities/availability.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+  id: string;
+
+  @Column({ name: 'date', type: 'timestamp' })
+  date: Date;
```
chelsea-EYDS, line 17: I feel like we will need a start date and an end date here

backend/src/common/enums/availability-type.enum.ts
fw-noel: Well I now know that AVAILABLE is default, so it should be UNAVAILABLE and DEPLOYED I think

backend/src/common/enums/experience.enum.ts
fw-noel: They would be, this is equivalent to
```
enum Experience {
INTERESTED = 'INTERESTED'
}
```

The thought is that we have an enum for what they should be stored in the database as (all caps, no spaces, underscores) and a different enum if we need the user friendly name for things. That latter enum for each thing might be frontend only, idk yet.

backend/src/database/entities/availability.entity.ts
fw-noel: Yep! I figured that there's no need to mask the ids for these, so I left UUID on personnel and ids on everything else

backend/src/database/entities/training.entity.ts
fw-noel: The latter! Personnel defines this
```
@ManyToMany(() => TrainingEntity)
  @JoinTable({ name: 'personnel_training' })
  trainings: TrainingEntity[];
```

backend/src/database/entities/availability.entity.ts
fw-noel: I thought it easier to define it day by day rather than a date range, but I'm open to discussion. This can be finalized when we're doing more of the availability / scheduling iscussions

backend/src/common/enums/availability-type.enum.ts
chelsea-EYDS: Should we set the default value automatically for a resource?

chelsea-EYDS: (we could also include this one later once we have fully spec'd out the scheduler.) I am not including availability on the dash yet since I see it as a dependency of the scheduler...

backend/src/common/enums/experience.enum.ts
chelsea-EYDS: cool makes sense ty!

backend/src/database/entities/availability.entity.ts
chelsea-EYDS: I vote not even adding in the availability until we do the scheduler. But I am also fine to include an update later.

backend/src/common/enums/availability-type.enum.ts
fw-noel: I'm not in full agreement that it should be a dependency of the scheduler, but it could certainly be after those discussions.

The way I have it now, we wouldn't have an entry for the user for the table if they are AVAILABLE since that is the default. On the frontend, we can just render it that way (days with missing values are AVAILABLE). To avoid from having an entry on this table for every single day of the month per user.

chelsea-EYDS (approved):

PR #49, opened 2024-01-24
[TC-235] request keycloak info from backend

fw-noel:
[TC-235](https://bcdevex.atlassian.net/browse/TC-235)

Objective:
While I'm open to doing this through devops, this for me is the path of least resistance.

[TC-235]: https://bcdevex.atlassian.net/browse/TC-235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/routes/AppRoutes.tsx
+++ b/frontend/src/routes/AppRoutes.tsx
@@ -1,6 +1,6 @@
 import { ReactKeycloakProvider } from '@react-keycloak/web';
 import Keycloak from 'keycloak-js';
-import { Suspense } from 'react';
+import { Suspense, useEffect, useState } from 'react';
 import { BrowserRouter, Route, Routes } from 'react-router-dom';
 import AppRoutes from './constants';
 import store from 'store';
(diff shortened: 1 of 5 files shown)


frontend/src/routes/AppRoutes.tsx
```diff
@@ -9,17 +9,32 @@ import { Dashboard, Login, NotFound, Profile } from '../pages';
...
 export default () => {
-  return (
-    <ReactKeycloakProvider
-      authClient={
-        new Keycloak({
-          realm: import.meta.env.VITE_KEYCLOAK_REALM,
-          url: import.meta.env.VITE_KEYCLOAK_AUTH_URL,
-          clientId: import.meta.env.VITE_KEYCLOAK_CLIENT,
-        })
+  const [keycloakInfo, setKeycloakInfo] = useState<Keycloak>();
```
chelsea-EYDS, line 15: This looks good. My only concern on this approach is having some sort of fallback page or handling the error from a UI perspective in case the request fails?

chelsea-EYDS: (not necessarily needed on this PR for approval, just something to consider in general I suppose)

chelsea-EYDS (approved):

frontend/src/routes/AppRoutes.tsx
chelsea-EYDS: Oh sorry I see that you have a fallback message. Maybe we can get Chevonne to design error/not found pages? Or maybe there is a fallback gov one we can borrow (but again - not for this PR)

fw-noel: Yes will need to look into options! An error page might be okay.

PR #52, opened 2024-01-26
[TC-228]: UI updates on dashboard

chelsea-EYDS:
[TC-0228](https://bcdevex.atlassian.net/browse/TC-0228)

Objective:

- Mostly just a bunch of minor styling changes
- Some change to the logic for location/region filters


frontend/src/routes/constants.ts
```diff
@@ -1,7 +1,7 @@
 export default Object.freeze({
   // Public routes
   Login: '/login',
-  Dashboard: '/dashboard',
+  Dashboard: '/',
```
fw-noel, line 4: I would suggest keeping this at Dashboard, and having work to automatically redirect an authenticated user to it. Remember, we will be having a splash page.

fw-noel:
BUG: If you select a function / exp and clear all, the column stays

Screenshot 2024-01-26 at 1 25 45 PM

frontend/src/routes/constants.ts
chelsea-EYDS: Okee, I thought the splash page would redirect a logged in user, and since we don't have any auth'd page at '/' I thought it made sense for the dashboard to be here but I can revert :)

chelsea-EYDS:
> BUG: If you select a function / exp and clear all, the column stays
> 
> Screenshot 2024-01-26 at 1 25 45 PM

Good catch!! TY! (I think it is resolved now)

fw-noel (approved):

PR #53, opened 2024-01-26
[TC-121] seed data + migrations (for local)

chelsea-EYDS:
[TC-0121](https://bcdevex.atlassian.net/browse/TC-0121)

Objective:

- use makefile cmd to generate seed data
- updated db module in order to run typeorm migrations
- added scripts to run migrations (locally for now - we can also keep syncing I just thought we should get these started ?)

### DIFF
--- a/backend/src/common/utils.ts
+++ b/backend/src/common/utils.ts
@@ -45,9 +45,8 @@ export const rowData = () => {
 };
 
 const experiences = () => {
-  const num = Math.floor(Math.random() * 9);
   const experiences = [];
-  for (let i = 0; i < num; i++) {
+  for (let i = 0; i < 5; i++) {
     const functionEntity: { id: number; name: string; abbreviation: string } =
       faker.helpers.arrayElement(functions);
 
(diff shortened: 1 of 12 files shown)


backend/src/common/utils.ts
```diff
@@ -58,4 +56,13 @@ export const generateData = () => {
...
+const handler = async () => {
+  const app = await NestFactory.createApplicationContext(AppModule);
+  const data = generateData();
+
+  const personnelService = app.get(PersonnelService);
+  console.log(data);
+  return await personnelService.createPersonnel(data);
+};
+
+console.log(handler());
```
fw-noel, line 68: ?

fw-noel (commented):
I'm not seeing when this is doing the seed? Is it on every startup?

backend/src/common/utils.ts
chelsea-EYDS: You have to run the makefile command and it only runs locally at this point. I just wanted to get an approval on this first. For seeding on dev/test we would only need to run once since the data persists, so I just figured I would tunnel in and run the makefile command.

fw-noel (dismissed):

fw-noel (approved):

PR #57, opened 2024-01-29
[FIX]: column style/display for func/exp on dash

chelsea-EYDS:

Objective:

- update styling to match designs for the table view for func/exp
Screenshot 2024-01-29 at 6 26 47 PM

### DIFF
--- a/frontend/src/components/filters/MultiSelect.tsx
+++ b/frontend/src/components/filters/MultiSelect.tsx
@@ -15,14 +15,21 @@ export const MultiSelect = ({
   values,
   label,
   onChange,
-  handleClose,
 }: {
   field: any;
   values: any;
   onChange: (e: ChangeEvent<HTMLInputElement>) => void;
-  handleClose: (name: string, value?: string) => void;
   label: string;
 }) => {
+  const handleClose = (name: string, value: string) => {
+    const event = {
+      target: {
+        name: name,
+        value: value,
+      },
+    } as ChangeEvent<HTMLInputElement>;
+    onChange(event);
+  };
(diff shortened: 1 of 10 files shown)


frontend/src/components/filters/MultiSelect.tsx
```diff
@@ -9,37 +9,88 @@ import {
...
-  handleClose,
 }: {
   field: any;
-  values: any;
+  values: string[];
   onChange: (e: ChangeEvent<HTMLInputElement>) => void;
-  handleClose: (name: string, value?: string) => void;
   label: string;
 }) => {
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
```
chelsea-EYDS, line 25: Moved these functions out of the useTable hook since they are specific to this component (and to avoid passing them in from the dash since it is not shared elsewhere)

frontend/src/components/filters/MultiSelect.tsx
```diff
@@ -50,21 +101,71 @@ export const MultiSelect = ({
...
-                  checked={values?.includes(option)}
-                  name={field.name}
-                  value={option}
-                />
-                {option}
-              </label>
-            </MenuItem>
-          ))}
-        </MenuList>
+        {field.groupedOptions ? (
```
chelsea-EYDS, line 104: combined MultiSelectGroup with MultiSelect because the logic for setting the values is shared, and only the MenuList view is different so this is now set conditionally

fw-noel (dismissed):

fw-noel (approved):

PR #62, opened 2024-01-30
[TC-236]: Sonarcloud config + fix warnings

chelsea-EYDS:
[TC-0236](https://bcdevex.atlassian.net/browse/TC-0236)

Objective:
- set up sonarcloud (not done in this PR - configured in the sonarcloud UI)
- addressed warnings

### DIFF
--- a/frontend/src/components/layout/Layout.tsx
+++ b/frontend/src/components/layout/Layout.tsx
@@ -10,8 +10,8 @@ export const Layout = ({
   username,
 }: {
   children: ReactElement;
-  authenticated?: boolean | undefined;
   username?: string;
+  authenticated?: boolean;
 }) => {
   return (
     <div className="min-h-screen flex flex-col">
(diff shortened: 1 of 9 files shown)


frontend/src/components/layout/Layout.tsx
```diff
@@ -10,8 +10,8 @@ export const Layout = ({
   username,
 }: {
   children: ReactElement;
-  authenticated?: boolean | undefined;
   username?: string;
+  authenticated?: boolean;
```
fw-noel, line 14: This is a sonarqube fix?
Couldn't we tell `authenticated` by the presence of a username?

fw-noel (approved):

frontend/src/components/layout/Layout.tsx
chelsea-EYDS: This isn't a new line, not sure why the whole line is showing as new. The fix was for a redundant | undefined since the authenticated is already set to optional with the "?" mark. 

Screenshot 2024-01-31 at 11 33 09 AM

PR #71, opened 2024-02-07
[FIX] fixed union membership flags

chelsea-EYDS:

Objective:
- Fixed the union membership buttons as I realized that I had included the new values in our mock data already so the new ones were unstyled.


chelsea-EYDS:
renamed

PR #77, opened 2024-02-08
[TC-174]: CHEFS integration - form json

chelsea-EYDS:
[TC-0174](https://bcdevex.atlassian.net/browse/TC-0174)

Objective:

 - Endpoint to receive webhook notification for a submitted form
 - Using the formId and SubmissionId to retrieve form data from CHEFS
 - saving form in a separate table (temporary) until we have final DTO (easier to retrieve form data and we can verify the requests are working properly)
 - token guard for validating the form ApiKey (bypasses auth guard with public decorator, but caught in token guard)


fw-noel (commented):
We should add documentation in the README for this piece

backend/src/auth/token_guard.ts
```diff
@@ -0,0 +1,41 @@
...
+    
+    const authToken = authHeader.split(' ')[1];
+
+    if (!authToken) {
+      throw new UnauthorizedException();
+    }
+    const token = process.env.CHEFS_WS_TOKEN;
+
+    if (!token) {
+      throw new UnauthorizedException();
```
fw-noel, line 26: This should probably be an Internal Server Error

backend/src/database/datasource.ts
```diff
@@ -10,7 +11,7 @@ export const config = {
   database: process.env.DB_NAME ?? 'tc',
   // TODO change this to false in production
   synchronize: true,
-  entities: [join(__dirname, '**', '*.entity.{ts,js}')],
+  entities: [Form, join(__dirname, '**', '*.entity.{ts,js}')],
```
fw-noel, line 14: Can we figure out if this is removable?

fw-noel (approved):

PR #79, opened 2024-02-12
[TC-279] BE changes for role permissions

chelsea-EYDS:
[TC-0279](https://bcdevex.atlassian.net/browse/TC-0279)

Objective:

- Use class validator to omit certain personnel fields if the user role is logistics (vs coordinator)

### DIFF
--- a/backend/src/personnel/ro/personnel.ro.ts
+++ b/backend/src/personnel/ro/personnel.ro.ts
@@ -100,7 +100,7 @@ export class PersonnelRO {
     required: false,
     example: 'MOTI Area Roads Mgr R24',
   })
-  @Expose({ groups: ['logistics'] })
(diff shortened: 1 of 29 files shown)


backend/src/personnel/ro/personnel.ro.ts
```diff
@@ -98,12 +95,20 @@ export class PersonnelRO {
   })
   skillsAbilities: string;
 
+  @ApiProperty({
+    description: 'Notes about Personnel - Only vieweable by coordinators',
+    required: false,
+    example: 'MOTI Area Roads Mgr R24',
+  })
+  @Expose({ groups: ['logistics'] })
```
chelsea-EYDS, line 103: fix

fw-noel (approved):

PR #80, opened 2024-02-12
[TC-276] FE changes for new user

chelsea-EYDS:
[TC-0276](https://bcdevex.atlassian.net/browse/TC-0276)

Objective:

- FE changes for displaying new users and role specific views


backend/src/auth/roles.guard.ts
```diff
@@ -45,6 +45,6 @@ export class RolesGuard implements CanActivate {
     } catch {
       throw new UnauthorizedException();
     }
-    return true;
```
fw-noel, line 48: Lol good catch

fw-noel (approved):

PR #96, opened 2024-02-21
[TC-258] Scheduler + Integration

fw-noel:
[TC-258](https://bcdevex.atlassian.net/browse/TC-258)

Objective:
Screenshot 2024-02-21 at 2 53 03 AM

[TC-258]: https://bcdevex.atlassian.net/browse/TC-258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/hooks/usePersonnel.tsx
+++ b/frontend/src/hooks/usePersonnel.tsx
@@ -1,14 +1,14 @@
 import { useEffect, useState } from 'react';
 import { AxiosPrivate } from '../utils';
-import type { Personnel } from '@/pages/dashboard';
+import type { Availability, Personnel } from '@/pages/dashboard';
 
 const usePersonnel = ({
   personnelId,
 }: {
   personnelId: string;
-}): { personnel: Personnel | undefined, availability: any, getAvailability: (from: string, to: string) => void } => {
+}): { personnel: Personnel | undefined, availability: Availability[], getAvailability: (from: string, to: string) => void } => {
   const [personnel, setPersonnel] = useState<Personnel>();
-  const [availability, setAvailability] = useState<any>();
+  const [availability, setAvailability] = useState<Availability[]>([]);
 
   useEffect(() => {
     (async () => {
--- a/frontend/src/hooks/usePersonnel.tsx
+++ b/frontend/src/hooks/usePersonnel.tsx
@@ -6,8 +6,9 @@ const usePersonnel = ({
   personnelId,
 }: {
   personnelId: string;
-}): { personnel: Personnel | undefined } => {
+}): { personnel: Personnel | undefined, availability: any, getAvailability: (from: string, to: string) => void } => {
   const [personnel, setPersonnel] = useState<Personnel>();
+  const [availability, setAvailability] = useState<any>();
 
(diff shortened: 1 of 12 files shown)


frontend/src/hooks/usePersonnel.tsx
```diff
@@ -16,8 +21,17 @@ const usePersonnel = ({
     })();
   }, [personnelId]);
 
+  const getAvailability = async (from: string, to: string) => {
+    const response = await AxiosPrivate.get(
```
chelsea-EYDS, line 25: Can we move this to the endpoint utils? (I just like having the list of all of the endpoints that we are using in one file - totally just a preference thing though if you feel otherwise :) )

frontend/src/hooks/usePersonnel.tsx
```diff
@@ -16,8 +21,17 @@ const usePersonnel = ({
...
+      `/personnel/${personnelId}/availability?from=${from}&to=${to}`,
```
chelsea-EYDS, line 26: I think we should be wrapping this in encodeURI() ?

frontend/src/pages/profile/SchedulerRow.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+const SchedulerRow = ({
+  month,
+  data,
+}: {
+  month: string;
+  data: SchedulerRowItem[];
+}) => {
+  // Because setting variable classes doesn't tend to work, we define all classes ahead of time
+  const availableStartClass =
```
chelsea-EYDS, line 83: Can we move this to a 'classes file'?

frontend/src/pages/profile/Scheduler.tsx
```diff
@@ -0,0 +1,148 @@
...
+  onChangeAvailabilityDates: (from: string, to: string) => void;
+}) => {
+  const [open, setOpen] = useState(1);
+  const handleOpen = (value: number) => setOpen(open === value ? 0 : value);
+  const [schedulerRows, setSchedulerRows] = useState<{
+    [key: string]: SchedulerRowItem[];
+  }>();
+
+  useEffect(() => {
+    if (availability) {
```
chelsea-EYDS, line 27: Can we set a default value for availability to avoid checking if exists?

chelsea-EYDS (approved):
Just a few nits but those could be addressed later. Looks great!

frontend/src/hooks/usePersonnel.tsx
fw-noel: Let's tag this on as tech debt, I think `useTable` has its own requests too.

frontend/src/pages/profile/Scheduler.tsx
fw-noel: Oh right, I forgot to take this out. We are defaulting to [] now from `usePersonnel`

chelsea-EYDS (dismissed):

chelsea-EYDS (approved):

PR #103, opened 2024-02-26
[TC-237] Backend changes

chelsea-EYDS:
[TC-0237](https://bcdevex.atlassian.net/browse/TC-0237)

Objective:

### DIFF
--- a/backend/src/personnel/personnel.service.ts
+++ b/backend/src/personnel/personnel.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
-import {  eachDayOfInterval, format, parse } from 'date-fns';
+import { eachDayOfInterval, format, parse } from 'date-fns';
 import { Brackets, Repository, UpdateResult } from 'typeorm';
 import { CreatePersonnelDTO } from './dto/create-personnel.dto';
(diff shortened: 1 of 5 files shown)


backend/src/personnel/personnel.service.ts
```diff
@@ -114,42 +114,48 @@ export class PersonnelService {
...
-          date: format(new Date(), 'yyyy-MM-dd'),
-        });
-      } else {
-        // This query is not very meaningful without a status - returns all personnel with any status within the date range - we shoudl enforce a status to be selecred if searching by date range
-        qb.andWhere('availability.date BETWEEN :from AND :to', {
-          from: query.availabilityFromDate,
-          to: query.availabilityToDate,
-        });
-      }
+    if (
```
fw-noel, line 133: I know these seem to be mutually exclusive, but I'd prefer to see else ifs here

chelsea-EYDS: I agree. I'm hoping to remove the "if no status but dates" check by enforcing a status since the query results are somewhat meaningless without one on the front end. I have a TODO above the second if chunk to update if that is sufficient for the moment?

fw-noel: Yeah sure

fw-noel (approved):

PR #104, opened 2024-02-26
[TC-273] FE changes

chelsea-EYDS:
[TC-0273](https://bcdevex.atlassian.net/browse/TC-0273)

Objective:

- Edit profile pop up form 
- field validation
*note: we may want to look at the field validation/dto's for creating/updating personnel a bit more closely. ie: shoudl we disable the email edit field if this is going to be a unique field in order to prevent duplicate form submissions from CHEFS

### DIFF
--- a/frontend/src/pages/profile/ProfileEditPopup.tsx
+++ b/frontend/src/pages/profile/ProfileEditPopup.tsx
@@ -10,7 +10,7 @@ import { EditProfileValidationSchema, fields, sections } from './constants';
 import { Divider } from '@/components/ui/Divider';
 import type { FormikHelpers, FormikProps, FormikState } from 'formik';
 import { Field, Form, Formik } from 'formik';
-import { Button, SectionHeader, Select, TextArea, TextInput } from '@/components';
+import { Button, SectionHeader, Select, TextInput } from '@/components';
 import { AxiosPrivate } from '@/utils';
 import { format } from 'date-fns';
 
@@ -87,7 +87,7 @@ export const ProfileEditPopup = ({
               placeholder={undefined}
               className="bg-grayBackground flex flex-row w-full justify-between"
             >
-              <span>Edit/Review Member Details</span>
+              <span>Edit Member Details</span>
               <button
                 className="text-sm text-primaryBlue underline font-normal"
                 onClick={handleOpenEditPopUp}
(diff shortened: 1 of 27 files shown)


frontend/src/pages/profile/ProfileEditPopup.tsx
```diff
@@ -0,0 +1,277 @@
...
+    );
+
+    Object.keys(personnel).forEach((key) => {
+      if (values[key as keyof Personnel] === personnel[key as keyof Personnel]) {
+        delete values[key as keyof Personnel];
+      }
+    });
+
+    try {
+      const res = await AxiosPrivate.patch(
```
fw-noel, line 59: I'd suggest adding this to `usePersonnel`, to keep the requests in a place.

frontend/src/pages/profile/ProfileEditPopup.tsx
```diff
@@ -0,0 +1,277 @@
...
+      default:
+        return false;
+    }
+  };
+  const handleSubmit = async (
+    values: Personnel,
+    helpers: FormikHelpers<Personnel>,
+  ) => {
+    // only send the fields that have been changed
+    values.remoteOnly = stringToBoolean(values.remoteOnly ? 'true' : 'false');
```
fw-noel, line 47: These string to booleans feel like they should be unnecessary.. Why are they necessary?

fw-noel (commented):
Some small comments.
Also, I noticed that when I click "Edit", the dialog starts a little bit scrolled down. 

Screenshot 2024-02-29 at 1 02 37 AM

Also, hitting "Edit" is minimizing the accordion. Nothing to do with fat fingering it. You'll want to use `stopPropagation` on the button, but watch that it also triggers `handleOpenEditPopUp` when you close the pop up so you might see some weird errors.

chelsea-EYDS:
> Some small comments. Also, I noticed that when I click "Edit", the dialog starts a little bit scrolled down.
> 
> Screenshot 2024-02-29 at 1 02 37 AM
> Also, hitting "Edit" is minimizing the accordion. Nothing to do with fat fingering it. You'll want to use `stopPropagation` on the button, but watch that it also triggers `handleOpenEditPopUp` when you close the pop up so you might see some weird errors.

ugh, I really don't like having the edit button inside of the button that opens the accordian. It's generating warnings in the console re: invalid nesting of buttons as well.

frontend/src/pages/profile/ProfileEditPopup.tsx
chelsea-EYDS: I'm using a Field from formik, the type for a value is "string|boolean|number". Even though the value is a boolean, when it's passed into the submit function it is a string, so I'm converting it to a boolean.

fw-noel (approved):

PR #105, opened 2024-02-27
[TC-282] Scheduler pop up, for add new event only

fw-noel:
[TC-282](https://bcdevex.atlassian.net/browse/TC-282)

Objective:
For add new event, scheduler pop up works
Some decisions have to be made for editing, so that will come later. The questions are:
- If I change a deployment from 03-01/03-10 to 03-03/03-10, what happens to 03-01 and 03-02? Back to Available?
- Same thing if the above block was Unavailable, 03-01/03-10. If I click on a block, I would expect that I'm making changes to the whole block since I'm displaying the whole range. However, if I'm just marking part of it (e.g. 03-09 and 03-10) as available / deployed, do I leave 03-01 til 03-08 as Unavailable?
- If I have a block where on the timeline it starts as the first day 03/01 because I only selected March, but the block actually started 02/25, what's the desired behavior for those February dates?

[TC-282]: https://bcdevex.atlassian.net/browse/TC-282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chelsea-EYDS:
> [TC-282](https://bcdevex.atlassian.net/browse/TC-282)
> 
> Objective: For add new event, scheduler pop up works Some decisions have to be made for editing, so that will come later. The questions are:
> 
> * If I change a deployment from 03-01/03-10 to 03-03/03-10, what happens to 03-01 and 03-02? Back to Available?
> * Same thing if the above block was Unavailable, 03-01/03-10. If I click on a block, I would expect that I'm making changes to the whole block since I'm displaying the whole range. However, if I'm just marking part of it (e.g. 03-09 and 03-10) as available / deployed, do I leave 03-01 til 03-08 as Unavailable?
> * If I have a block where on the timeline it starts as the first day 03/01 because I only selected March, but the block actually started 02/25, what's the desired behaviour for those February dates? 

Hey I was thinking about the "what happens to x when y is changed" q's and one approach might be just to leave them as they are. What I mean is that if someone wants to change a deployment from 03-01/03-10 to 03-03/03-10, then the action would not be to change the start date but instead to change those two days to whatever the new status is. If the resource will be available, they should set to available. If the deployment has changed to start on the third and the resource has not indicated their availability then it should revert to NOT_INDICATED (which is the default so I think that makes more sense than available).

frontend/src/pages/profile/ProfileHeader.tsx
```diff
@@ -23,7 +23,7 @@ const ProfileHeader = ({
   role,
 }: {
   personnel: Personnel;
-  role: Role;
+  role: Role | undefined;
```
chelsea-EYDS, line 26: `role?: Role` ?

chelsea-EYDS (approved):

PR #106, opened 2024-02-28
[TC-255]: work locations, regions, functions updated

chelsea-EYDS:
[TC-0255](https://bcdevex.atlassian.net/browse/TC-0255)

Objective:

- ensure all regions/location/function lists are up to date
- add location/region list to db
- populate dashboard fields with the values returned from the BE


backend/src/database/entities/personnel.entity.ts
```diff
@@ -36,15 +38,15 @@ export class PersonnelEntity extends BaseEntity {
...
-    type: 'varchar',
-    length: 100,
-  })
-  workLocation: string;
-
-  @Column({ name: 'region', type: 'enum', enum: Region, enumName: 'region' })
-  region: Region;
+  @JoinColumn([
+    {
+      name: 'workLocation',
```
fw-noel, line 43: Can we keep db column names with underscores?
And since this is a reference, I think `work_location_id` would be more proper

fw-noel (approved):

PR #109, opened 2024-03-02
[FIX] dash filters data

chelsea-EYDS:

Objective:

- fix the request to get dashboard filters

### DIFF
--- a/frontend/src/pages/dashboard/constants.ts
+++ b/frontend/src/pages/dashboard/constants.ts
@@ -42,7 +42,7 @@ export const dashboardToggle = {
   name: DashboardFilterNames.SHOW_INACTIVE,
   label: 'Show Inactive',
 };
-export interface LocationInterface {
+export interface Location {
   id: string;
   locationName: string;
   region: Region;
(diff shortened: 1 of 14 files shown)


frontend/src/pages/dashboard/constants.ts
```diff
@@ -42,7 +42,7 @@ export const dashboardToggle = {
   name: DashboardFilterNames.SHOW_INACTIVE,
   label: 'Show Inactive',
 };
-export interface WorkLocation {
+export interface LocationInterface {
```
fw-noel, line 45: Why this change? We don't tend to name our interfaces `xInterface` (I see now that there's a few others that I missed), so curious as to why here!

fw-noel (dismissed):

frontend/src/pages/dashboard/constants.ts
chelsea-EYDS: WorkLocation was already an enum, and then I wanted to make it more generic since we're using this for both home and work, and then I just didn't remove the Interface. I add interface when there is a naming conflict. I supposed I could just alias the import but it makes searching easier if they're named differently. I will fix :)

fw-noel (approved):

PR #113, opened 2024-03-04
[TC-355] Availability Filter 

chelsea-EYDS:
[TC-0355](https://bcdevex.atlassian.net/browse/TC-0355)

Objective:

- instead of partial, show # of # days for a partial result
- return all resources on page load, include availability status if exists (for the current date) or NOT_INDICATED otherwise
- ensure page load is quick

### DIFF
--- a/frontend/src/components/table/classes.ts
+++ b/frontend/src/components/table/classes.ts
@@ -50,6 +50,7 @@ const getUnionMembershipClass = (value?: string) => {
 };
 
 export const getAvailabilityClass = (value?: string) => {
+
   switch (value) {
     case AvailabilityTypeName.AVAILABLE:
       return tableClasses.available;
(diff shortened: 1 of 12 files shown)


frontend/src/components/table/classes.ts
```diff
@@ -68,7 +68,7 @@ export const tableClass = (key: string, value?: string) => {
     case DashboardColumns.REMOTE:
       return tableClasses.remote;
     case DashboardColumns.AVAILABILITY:
-      return getAvailabilityClass(value);
+      return getAvailabilityClass(value?.split(' ')[0]);
```
fw-noel, line 71: I don't like this, but I don't see an easy way around

fw-noel (approved):

frontend/src/components/table/classes.ts
chelsea-EYDS: I know. It's not great.

fw-noel (approved):

PR #120, opened 2024-03-06
[TC-343] Edit Notes

chelsea-EYDS:
[TC-0343](https://bcdevex.atlassian.net/browse/TC-0343)

Objective:
- edit notes

### DIFF
--- a/frontend/src/pages/profile/Profile.tsx
+++ b/frontend/src/pages/profile/Profile.tsx
@@ -22,6 +22,9 @@ import type { AvailabilityRange } from '../dashboard';
 import { ProfileEditForm } from './ProfileEditForm';
 import type { AvailabilityType } from '@/common';
 import ProfileFunctions from './ProfileFunctions';
+import ProfileNotes from './ProfileNotes';
+import { EditNotes } from './EditNotes';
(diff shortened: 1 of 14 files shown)


frontend/src/pages/profile/Profile.tsx
```diff
@@ -112,20 +127,61 @@ const Profile = () => {
...
+              handleOpenEditNotes={handleOpenEditNotes}
+              handleOpenEditCoordinatorNotes={handleOpenEditCoordinatorNotes}
+            />
+            <DialogUI
+              open={openEditNotes}
+              onClose={handleOpenEditNotes}
+              handleOpen={handleOpenEditNotes}
+              title={'Edit Notes'}
+            >
+              <EditNotes
```
fw-noel, line 161: I feel like this could be simplified a bit more, ie. instead of two <EditNotes> to have one that accepts which field is to be edited.

I'm not fussed though. However, can you move the <DialogUI> components outside of this div? In the same part of the tree that the <Dialog>s are in?

fw-noel (dismissed):

fw-noel (approved):

PR #126, opened 2024-03-07
[TC-369] bug fix availability filter on dash

chelsea-EYDS:
[TC-0369](https://bcdevex.atlassian.net/browse/TC-0369)

Objective:

### DIFF
--- a/backend/src/common/helpers.ts
+++ b/backend/src/common/helpers.ts
@@ -1,6 +1,7 @@
 /**
  *
  * @param date Used when calling new Date() on the backend to query for availability events saved with a PST date string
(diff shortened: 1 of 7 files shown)


backend/src/common/helpers.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ *
+ * @param date Used when calling new Date() on the backend to query for availability events saved with a PST date string
```
fw-noel, line 3: Can you add a comment here to avoid using this when SAVING dates?

chelsea-EYDS: For sure.

fw-noel (approved):

PR #136, opened 2024-03-11
[TC-0384]: fix favicon

chelsea-EYDS:
[TC-0384](https://bcdevex.atlassian.net/browse/TC-0384)

Objective:

- aded manifest to vite config


fw-noel (dismissed):
This feels like a lot of changes for a small issue :P but I believe ya

fw-noel (approved):

PR #139, opened 2024-03-11
[TC-364] fill in empty date for "today"

fw-noel:
[TC-364](https://bcdevex.atlassian.net/browse/TC-364)

Objective:
This isn't 100% related to TC-364, just that it was the reason it was sent back.
Basically, for personnel that do not have a date for today, it was still pulling in a lot of their dates, and it wasn't necessarily reflective of their NOT_INDICATED status.
The change here is to make it so that personnel.availability is rightly empty, and then backfilled later if it is empty. This might have repercussions in the future if we're pulling a single entity, so we may later want to move this logic to the GET() endpoint itself, or figure out another way.

[TC-364]: https://bcdevex.atlassian.net/browse/TC-364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/src/database/entities/personnel.entity.ts
+++ b/backend/src/database/entities/personnel.entity.ts
@@ -15,11 +15,13 @@ import { LocationEntity } from './location.entity';
 import { ExperienceEntity } from './personnel-function-experience.entity';
 import { TrainingEntity } from './training.entity';
 import { Role } from '../../auth/interface';
-import { Status } from '../../common/enums';
+import { AvailabilityType, Status } from '../../common/enums';
 import { Classification } from '../../common/enums/classification.enum';
 import { Ministry } from '../../common/enums/ministry.enum';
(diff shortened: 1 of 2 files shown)


chelsea-EYDS:
> [TC-364](https://bcdevex.atlassian.net/browse/TC-364)
> 
> Objective: This isn't 100% related to TC-364, just that it was the reason it was sent back. Basically, for personnel that do not have a date for today, it was still pulling in a lot of their dates, and it wasn't necessarily reflective of their NOT_INDICATED status. The change here is to make it so that personnel.availability is rightly empty, and then backfilled later if it is empty. This might have repercussions in the future if we're pulling a single entity, so we may later want to move this logic to the GET() endpoint itself, or figure out another way.

This is great, good solution.

PR #171, opened 2024-03-18
[FIX] dynamic form url

chelsea-EYDS:

Objective:
- set the url for chefs form dynamically


backend/src/app.controller.ts
```diff
@@ -68,4 +68,14 @@ export class AppController {
       locations: await this.locationService.getRegionsAndLocations(),
     };
   }
+/**
+ * Returns the formId for the chefs form based on the current env
+ * @returns 
+ */
+  @Get('/formId')
```
fw-noel, line 75: /form/id or just simply /form please

fw-noel (approved):

PR #173, opened 2024-03-18
[TC-408] FIX loading component + rename work/home location

chelsea-EYDS:
[TC-0408](https://bcdevex.atlassian.net/browse/TC-0408)

Objective:


fw-noel (approved):
A bit strange, but sure. Why not just put the whole table in the conditional?

chelsea-EYDS:
> A bit strange, but sure. Why not just put the whole table in the conditional?

I tried that first. It placed it in the first column, so it was still stuck on the left.

PR #176, opened 2024-03-19
[TC-124] file rotation saves for info and separate for errors

fw-noel:
[TC-124](https://bcdevex.atlassian.net/browse/TC-124)

Objective:
For now, we'll do file rotation saves. We have a separate one for logging that clears after 2 weeks and one for errors that don't, but rotates monthly. We can tweak this as we see use; if there's too much logging or what have you.

[TC-124]: https://bcdevex.atlassian.net/browse/TC-124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -41,7 +41,7 @@
     "@nestjs/platform-express": "10.0.0",
     "@nestjs/swagger": "7.1.16",
     "@nestjs/terminus": "10.2.0",
-    "@nestjs/throttler": "^5.1.1",
+    "@nestjs/throttler": "5.1.1",
     "@nestjs/typeorm": "10.0.1",
     "axios": "1.6.2",
     "class-transformer": "0.5.1",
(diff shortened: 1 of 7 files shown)


backend/package.json
```diff
@@ -55,7 +55,8 @@
     "reflect-metadata": "0.1.13",
     "rxjs": "7.8.1",
     "ts-node": "10.9.1",
-    "typeorm": "0.3.17"
+    "typeorm": "0.3.17",
+    "winston-daily-rotate-file": "^5.0.0"
```
chelsea-EYDS, line 59: I don't know if it matters, but should we lock the version?

chelsea-EYDS (approved):

chelsea-EYDS (approved):

PR #181, opened 2024-03-20
[TC-420] added new locations to the seed script

chelsea-EYDS:
[TC-0420](https://bcdevex.atlassian.net/browse/TC-0420)

Objective:

- added new locations to the seed script


fw-noel (commented):
Do we need it in common/utils and the ENUM as well?

fw-noel (approved):

PR #199, opened 2024-04-09
[TC-458] add ICS training for the backend

fw-noel:
[TC-458](https://bcdevex.atlassian.net/browse/TC-458)

Objective:
Migration for Approved by Supervisor
Seeding functions for Trainings
Returning both values to the frontend

[TC-458]: https://bcdevex.atlassian.net/browse/TC-458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/src/database/entities/personnel.entity.ts
+++ b/backend/src/database/entities/personnel.entity.ts
@@ -24,7 +24,7 @@ import { UnionMembership } from '../../common/enums/union-membership.enum';
 import { datePST } from '../../common/helpers';
 import { CreatePersonnelDTO } from '../../personnel/dto/create-personnel.dto';
 import { PersonnelRO } from '../../personnel/ro/personnel.ro';
-import { ICS_TRAINING_NAME } from 'src/common/const';
+import { ICS_TRAINING_NAME } from '../../common/const';
 
 @Entity('personnel')
 export class PersonnelEntity extends BaseEntity {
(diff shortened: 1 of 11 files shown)


backend/src/database/entities/personnel.entity.ts
```diff
@@ -24,6 +24,7 @@ import { UnionMembership } from '../../common/enums/union-membership.enum';
 import { datePST } from '../../common/helpers';
 import { CreatePersonnelDTO } from '../../personnel/dto/create-personnel.dto';
 import { PersonnelRO } from '../../personnel/ro/personnel.ro';
+import { ICS_TRAINING_NAME } from 'src/common/const';
```
chelsea-EYDS, line 27: I think this will fail somehow (the import)

backend/src/personnel/personnel.service.ts
```diff
@@ -56,7 +59,8 @@ export class PersonnelService {
     });
 
     try {
-      await this.personnelRepository.update(id, { ...person });
+      // This is a 'save' rather than 'update' to allow for updating many-to-many relations
```
chelsea-EYDS, line 62: Just want to double check that we're not passing in a partial "person" here on save?

chelsea-EYDS (approved):
Looks good! (Please see comment re: partial person though - I know on the FE we are only sending the data to update, not the full personnel, so I just wanted to make sure that we don't save over the person with partial data... if that makes sense?)

backend/src/personnel/personnel.service.ts
fw-noel: I don't believe it could be a partial person, since we pull it out of the repository first and replace each property that's changed

chelsea-EYDS: Right, sorry I could have looked at that :P

PR #202, opened 2024-04-13
[TC-490] colours

chelsea-EYDS:
[TC-0490](https://bcdevex.atlassian.net/browse/TC-0490)

Objective:
- fix bug caused by tailwind config colours update
- fixed error from using Date | string on personnel interface
- fixed local seed script


fw-noel (approved):

chelsea-EYDS:
merging (previously approved and Noel is on vacation)

PR #216, opened 2024-04-27
[TC-0502]: changes to roles guard and token guard

chelsea-EYDS:
[TC-0502](https://bcdevex.atlassian.net/browse/TC-0502)

Objective:

- added new enum 'program' (refers to BCWS, EMCR or ADMIN)

- added a new token to be verified by the token guard

- auth guard, roles guard, token guard are set globally, so each request must pass each guard

- auth guard to assign program to the request object (if auth'd)

     - keycloak client_roles may contain any of: 'bcws', 'emcr', 'logistics', 'coordinator'
if client_roles contains both of 'bcws' && 'emcr' then program 'admin' is assumed
in the case of admin (access to both bcws and emcr) the coordinator role is assumed in the case that both 'logistics' and 'coordinator' are present.
(the keycloak client roles don't have a lot of granularity - for example, in the case of: ['bcws', 'emcr', 'logistics', 'coordinator'] we cannot know whether logistics applies to bcws or emcr.)

- if a route is annotated with @ public() it will pass through all guards

- if a route is annotated with @ token() it will be passed by the auth guard (auth guard requires a valid bearer token, whereas token guard is used for integration with CHEFS and e-diaries and uses an ApiKey)

- if a route is not annotated with @ public() or @ token() then the request must contain a valid bearer token. The auth guard will verify the token and then deconstruct the client_roles from the token user object.

- client_roles are assigned to the request object, and must contain one of 'bcws' or 'emcr' or both (as program), as well as one of 'logistics' or 'coordinator' or both (as role)

- auth guard assigns request.program = 'bcws'|'emcr'|'admin'

- auth guard assigns request.role = 'coordinator'|'logistics' (if both are present coordinator takes precedence)

- after a request passes the auth guard verification and is assigned the relevant program/role it is passed to the RolesGuard.

- if no required roles have been included in the Roles() decorator on the request route then the request will pass through

- if a route contains the Roles decorator then the request.role must be present in the required roles (set by the endpoint decorator and retrieved by the reflector metadata)

- requests annotated with @ token will require a valid token (set by the endpoint decorator and retrieved by the reflector metadata)


backend/src/auth/interface.ts
```diff
@@ -41,6 +52,7 @@ interface ResourceAccess {
 }
 
 export interface RequestWithRoles extends Request {
-  role: Role;
+  role?: Role;
+  program: Program[];
```
fw-noel, line 56: Why array, given that `admin` is used for both?

backend/src/bcws/bcws.controller.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Programs } from '../auth/program.decorator';
+import { Token } from '../auth/token.decorator';
+
+import { AppLogger } from '../logger/logger.service';
+
+@ApiTags('BcwsController API')
+@UseInterceptors(ClassSerializerInterceptor)
+@Programs([Program.BCWS])
+@Controller('bcws')
+export class BcwsController {
```
fw-noel, line 20: I don't love this having its own controller

fw-noel:
Will be testing tomorrow, but made comments :)

backend/src/bcws/bcws.controller.ts
chelsea-EYDS: I thought we would need a controller for BCWS anyways since we will have BCWS specific endpoints for the dashboard and the profile as well. I agree on not having a controller for a single endpoint though.

backend/src/auth/auth.guard.ts
```diff
@@ -107,20 +129,55 @@ export class AuthGuard implements CanActivate {
     }
   }
 
-  setDevRoles(payload: JwtPayload, request: Request): void {
-    // only include a single role in the request object.
-    // include the role with the highest permissions level.
+  setDevProgramRoles(payload: JwtPayload, request: Request): void {
+    const hasValidProgramRole =
```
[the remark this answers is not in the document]
chelsea-EYDS: Good call

backend/src/auth/interface.ts
chelsea-EYDS: Why, indeed? Will change that, it does not need to be an array

.env.example
```diff
@@ -18,3 +18,13 @@ API_VERSION=0.0.1
 
 ENV=ci
 TARGET=dev
+
+
+
+CHEFS_API=https://submit.digital.gov.bc.ca
```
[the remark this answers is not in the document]
chelsea-EYDS: Agreed - except that it already exists throughout the code and on OC as CHEFS_API so will require more updates. (Which can be done, just wanted to make sure it is worthwhile to rename?)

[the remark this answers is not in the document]
fw-noel: Nah :D

backend/src/auth/auth.guard.ts
```diff
@@ -25,18 +31,33 @@ export class AuthGuard implements CanActivate {
       return true;
     }
 
+    const TokenEndpoint = this.reflector.getAllAndOverride<boolean>(
```
fw-noel, line 34: Why is `PublicEndpoint` and `TokenEndpoint` capitalized?

backend/src/auth/token.guard.ts
```diff
@@ -0,0 +1,83 @@
...
+      Metadata.TOKEN_TYPE,
+      context.getHandler(),
+    );
+
+    if (!requiredTokens) {
+      return true;
+    }
+
+    const request = context.switchToHttp().getRequest();
+    const authHeader = request.headers.authentication;
```
fw-noel, line 43:
```suggestion
    const authHeader = request.headers.authorization;
```

backend/src/auth/token.guard.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  validateChefsToken(authHeader: string) {
+    if (authHeader !== process.env.CHEFS_WS_TOKEN) {
+      this.logger.error('Unauthorized - invalid auth token');
+      throw new UnauthorizedException();
+    }
+  }
+  validateBcwsToken(authHeader: string) {
```
fw-noel, line 77: These are still including `BEARER `, if we're asking them to send such a token. I can't remember what we said we're expecting them to send; I know it was a single value, but I can't remember if BEARER or not

fw-noel (changes requested):

backend/src/auth/token.guard.ts
chelsea-EYDS: I did this as "authorization" b/c the CHEFS docs used "authorization" for their key. But, I think we can rename the key, I had just followed the example from their documentation.

fw-noel (approved):

PR #222, opened 2024-05-03
[TC-500] BCWS Entities

chelsea-EYDS:
[TC-0500](https://bcdevex.atlassian.net/browse/TC-0500)

Objective:

- created bcws entities/enums/ro/dto


backend/src/database/entities/bcws/bcws-personnel.entity.ts
```diff
@@ -0,0 +1,219 @@
...
+  @Column({ name: 'purchase_card_holder', type: 'boolean', default: false })
+  purchaseCardHolder: boolean;
+
+  //TODO - add this once we have the values for the branch enum
+  // @Column({ name: 'branch', type: 'enum', enum: Branch })
+  // branch: Branch;
+
+  //TODO confirm length of paylist_id
+  @Column({ name: 'paylist_id', type: 'varchar', length: 6 })
+  paylistId: number;
```
fw-noel, line 67: What are these?

backend/src/database/entities/bcws/bcws-personnel.entity.ts
```diff
@@ -0,0 +1,219 @@
...
+    enumName: 'ofa',
+    nullable: true,
+  })
+  highestOFA?: OFA;
+
+  @Column({ name: 'ofa_expiry', type: 'timestamp', nullable: true })
+  ofaExpiryDate?: Date;
+
+  @Column({ name: 'food_safety_I', type: 'boolean', default: false })
+  foodSafetyLevelI?: boolean;
```
fw-noel, line 127: I forget; did we say we were going to move this out of this table?

backend/src/database/entities/bcws/bcws-personnel.entity.ts
```diff
@@ -0,0 +1,219 @@
...
+    const personnelData = this.personnel.toResponseObject(role, lastDeployed);
+    const data = {
+      ...personnelData,
+      homeFireCentre: this?.homeFireCentre?.toResponseObject() ?? {},
+      workFireCentre: this?.workFireCentre?.toResponseObject() ?? {},
+      employeeId: this.employeeId,
+      paylistId: this.paylistId,
+      dateApplied: this.dateApplied,
+      dateApproved: this.dateApproved,
+      coordinatorNotes: this.coordinatorNotes,
```
fw-noel, line 182: Wondering if there's a scenario that will warrant renaming this to `bcwsCoordinatorNotes` and `bcwsLogisticsNotes`. Whether just in the RO or the entity itself. Not seeing it immediately, so if you don't see it, feel free to disregard this comment.

backend/src/database/queries.ts
```diff
@@ -80,4 +80,136 @@ export const functionSql = `INSERT INTO public."function" (name,abbreviation) VA
 	 ('Planning','Plans'),
 	 ('Advanced Planning Unit','APU'),
 	 ('Recovery','Recovery'),
-	 ('Deputy Director','DDir');`;
\ No newline at end of file
+	 ('Deputy Director','DDir');`;
+
+export const bcwsLocations = `
```
fw-noel, line 85:
```suggestion
export const bcwsLocationsSql = `
```

backend/src/database/queries.ts
```diff
@@ -80,4 +80,136 @@ export const functionSql = `INSERT INTO public."function" (name,abbreviation) VA
...
+   (50, 'Cranbrook','SOUTHEAST'),
+   (51, 'Kimberley','SOUTHEAST'),
+   (52, 'Nelson','SOUTHEAST'),
+   (53, 'Nakusp','SOUTHEAST'),
+   (54, 'Castelgar','SOUTHEAST'),
+   (55, 'Golden','SOUTHEAST'),
+   (56, 'Grand Forks','SOUTHEAST');
+   `;
+
+export const roles = `INSERT INTO public.bcws_role (id,"name","section") VALUES
```
fw-noel, line 145:
```suggestion
export const bcwsRolesSql = `INSERT INTO public.bcws_role (id,"name","section") VALUES
```

fw-noel: comment applies to the below sql statements too

backend/src/database/seed-certifications.ts
```diff
@@ -0,0 +1,16 @@
+import { datasource } from './datasource';
+import { certs } from './queries';
+
+const seedLocation = async () => {
```
fw-noel, line 4: Am I missing when these are actually going to run?

backend/src/personnel/dto/bcws/create-bcws-personnel.dto.ts
```diff
@@ -0,0 +1,199 @@
...
+  @ApiProperty({
+    description: 'Ministry personnel works in',
+    enum: FireCentre,
+    example: FireCentre.CARIBOO,
+  })
+  @IsEnum(FireCentre)
+  fireCentre?: FireCentre;
+}
+
+export class CreatePersonnelBcwsDTO {
```
fw-noel, line 33: Should this extend PersonnelDTO?

backend/src/database/entities/bcws/bcws-personnel.entity.ts
chelsea-EYDS: I was torn on this one. If we use the same key name for the values which have the same key for both tables, but different values we can just keep them in the personnelRO which each program extends, which I think will result in less changes for the current places these keys are referenced. However, I could also see a case for renaming the keys of these fields.

backend/src/database/seed-certifications.ts
chelsea-EYDS: Yes, they will run in the TC-500-personnel PR. Just trying to split these up for readability without breaking the tests.

backend/src/personnel/dto/bcws/create-bcws-personnel.dto.ts
chelsea-EYDS: no, we'll want to create a single person, and then create bcws/emcr data (otherwise if they both extend person and we call create on each of them for someone applying to both programs we'll have two personnel entities)

backend/src/database/entities/bcws/bcws-personnel.entity.ts
fw-noel: We can keep for now, easy to change in the future

backend/src/personnel/dto/bcws/create-bcws-personnel.dto.ts
fw-noel: I follow for now. May want to change in the future.

backend/src/database/seed-certifications.ts
fw-noel: Last one, this should be `seedBcwsCertifications` and the other seed should be `seedFireCentreLocations`

fw-noel (approved):

PR #223, opened 2024-05-03
[TC-500] EMCR entities

chelsea-EYDS:
[TC-0500](https://bcdevex.atlassian.net/browse/TC-0500)

Objective:

- created emcr entities/enums/ro/dto (copied from pre-existing fields on personnel)


fw-noel (approved):
Looks fine, but lets not merge until the others are addressed

PR #239, opened 2024-05-11
[TC-544] Dashboard Filters

chelsea-EYDS:
[TC-0544](https://bcdevex.atlassian.net/browse/TC-0544)

Objective:

- Dashboard filters + columns


backend/src/database/migrations/1715451072374-division-migration.ts
```diff
@@ -0,0 +1,29 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class Migration1715451072374 implements MigrationInterface {
```
fw-noel, line 3:
```suggestion
export class DivisionMigration1715451072374 implements MigrationInterface {
```

backend/src/personnel/dto/bcws/create-bcws-personnel.dto.ts
```diff
@@ -170,4 +166,10 @@ export class CreatePersonnelBcwsDTO {
   })
   @IsArray()
   certifications: CreateBcwsCertificationsDTO[];
+
+  @ApiProperty({
+    description: 'Division and Ministry',
+    required: false,
+  })
+  division: DivisionEntity;
```
fw-noel, line 174: Using an entity in a DTO is quite strange. I like the division id way

backend/src/personnel/personnel.service.ts
```diff
@@ -318,25 +349,39 @@ export class PersonnelService {
...
 
-    if (status === Status.PENDING) {
-      queryBuilder.orderBy('bcws_personnel.dateApplied', 'ASC');
-      queryBuilder.addOrderBy('personnel.lastName', 'ASC');
-      queryBuilder.addOrderBy('personnel.firstName', 'ASC');
-    } else {
-      queryBuilder.orderBy('bcws_personnel.dateApproved', 'DESC');
-      queryBuilder.addOrderBy('personnel.lastName', 'ASC');
-      queryBuilder.addOrderBy('personnel.firstName', 'ASC');
+    if (tableName === 'bcws_personnel') {
```
fw-noel, line 363: Ack, good catch!
Can you use `program` from the param instead of `tableName`

frontend/src/components/table/body/TableBodyCell.tsx
```diff
@@ -34,39 +46,152 @@ export const TableBodyCell = ({
...
+              <MinusIcon className={iconClass(cell.value)} /> Pending
+            </td>
+          )}
+        </>
+      );
+    case DashboardColumns.LOCATION:
+      return <td className={tableClasses.default + ' text-left'}>{cell.value}</td>;
     default:
-      return <td className={cell.className}>{cell.value}</td>;
+      return <td className={tableClasses.default + ' text-left ml-4'}>{cell.value}</td>;
```
fw-noel, line 195: Any reason you're not using `${template literals}`?

frontend/src/components/ui/Table.tsx
```diff
@@ -1,32 +1,176 @@
...
+          </tr>
+        )}
+        <tr>
+          <th
+            className={`sticky h-[10px] bg-white top-[226px] border-b-1 border-slate-100`}
+            colSpan={columns.length}
+          ></th>
+        </tr>
+        <tr
+          className={`sticky px-6  bg-white  h-[48px] shadow-blue-gray-900 ${role === Role.COORDINATOR ? ' top-[225px]' : 'top-40'}`}
```
fw-noel, line 151: Historically, I've seen tailwind have issues with conditional classNames like this, which is why I've had to define two different classNames in their own strings, and use the defined strings within the conditional. Are you not seeing issues here?

frontend/src/hooks/helpers.ts
```diff
@@ -51,188 +48,208 @@ export const getAvailabilityValue = (
...
-        DashboardColumns.REGION,
-        personnel.homeLocation?.region?.toLowerCase() ?? '',
-      ),
+      value: personnel.homeLocation?.region,
+    },
+
+    fireCentre: {
+      key: DashboardColumns.FIRE_CENTRE,
+      columnName: DashboardColumns.FIRE_CENTRE,
+      value: FireCentreName[personnel?.homeFireCentre?.fireCentre],
```
fw-noel, line 98: Here, but also in general for this file, since personnel and homeFireCentre are nullished, do we need to define a value if anything is null?
i.e.
`FireCentreName[personnel?.homeFireCentre?.fireCentre] || 'Unknown'`

frontend/src/hooks/useGetFilters.tsx
```diff
@@ -6,56 +6,96 @@ import {
...
   const [functions, setFunctions] = useState<FunctionType[]>([]);
+  const { route } = useRole();
   const { AxiosPrivate } = useAxios();
 
   const sortRegion = (reg: Region[]) => {
     return reg.sort((a: Region, b: Region) => a.localeCompare(b));
   };
 
+  const sortFireCentre = (centre: FireCentre[]) => {
+    return centre.sort((a: FireCentre, b: FireCentre) => a.localeCompare(b));
```
fw-noel, line 32: Would this fit better in the backend?

frontend/src/hooks/useTable.tsx
```diff
@@ -36,26 +36,45 @@ export const useTable = (route: Program | undefined) => {
   const tabs = [
     { index: 0, label: StatusNames.ACTIVE, value: Status.ACTIVE },
     { index: 1, label: StatusNames.INACTIVE, value: Status.INACTIVE },
-    { index: 2, label: StatusNames.PENDING, value: Status.PENDING },
+    { index: 2, label: 'Pending Approval', value: Status.PENDING },
```
fw-noel, line 39: Eh?

frontend/src/hooks/useGetFilters.tsx
chelsea-EYDS: I don't think so - we are returning locations list in a proper order, and then just re-ordering the associated fire centres for displaying in the drop down. If we order by fire centre then locations will be out of order.

frontend/src/hooks/useTable.tsx
chelsea-EYDS: It's just the label being changed in one place (as per figma) I didn't want to change the enum since "Pending" is  correct, except that on the tabs it is displayed differently now.

frontend/src/components/ui/Table.tsx
chelsea-EYDS: am not having any issues - I think before it might have been b/c we were passing in the value, rather than setting it conditionally inline? (just a guess)

backend/src/personnel/personnel.service.ts
```diff
@@ -219,13 +233,23 @@ export class PersonnelService {
...
+      query.availabilityFromDate,
+      query.availabilityToDate,
+    );
 
     if (query.fireCentre?.length) {
-      qb.andWhere('bcws_personnel.homeFireCentre.fireCentre IN (:...fireCentres)', {
-        fireCentres: query.fireCentre,
-      });
+      qb.andWhere(
+        'bcws_personnel.homeFireCentre.fireCentre IN (:...fireCentres)',
```
fw-noel, line 248:
```suggestion
        'homeFireCentre.fireCentre IN (:...fireCentres)',
```

Same change for line 256 please. That's on me

fw-noel (approved):

PR #240, opened 2024-05-13
[TC-554] get bcws

fw-noel:
[TC-554](https://bcdevex.atlassian.net/browse/TC-554)

Objective:
Need to fix what this branch is pointing to

[TC-554]: https://bcdevex.atlassian.net/browse/TC-554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/src/database/entities/bcws/bcws-personnel-roles.entity.ts
+++ b/backend/src/database/entities/bcws/bcws-personnel-roles.entity.ts
@@ -28,11 +28,19 @@ export class BcwsSectionsAndRolesEntity {
   })
   expLevel: ExperienceLevel;
 
+  @Column({
+    name: 'rank',
+    type: 'int',
+    default: 1,
+  })
+  rank: number;
+
   toResponseObject(): BcwsPersonnelRoleRO {
     return {
       role: this.role.name,
       section: this.role.section,
       expLevel: this.expLevel,
+      rank: this.rank,
     };
   }
 }
--- a/backend/src/database/entities/bcws/bcws-personnel-roles.entity.ts
+++ b/backend/src/database/entities/bcws/bcws-personnel-roles.entity.ts
@@ -31,7 +31,6 @@ export class BcwsSectionsAndRolesEntity {
   @Column({
     name: 'rank',
     type: 'int',
-    default: 1,
   })
   rank: number;
 
--- a/backend/src/database/migrations/1715585079303-migration.ts
+++ b/backend/src/database/migrations/1715585079303-migration.ts
@@ -1,0 +1,22 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Migration1715585079303 implements MigrationInterface {
+    name = 'Migration1715585079303'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "bcws_personnel_roles" ADD "rank" integer NOT NULL`);
+        await queryRunner.query(`ALTER TABLE "bcws_personnel" ADD "emergency_contact_first_name" character varying(50)`);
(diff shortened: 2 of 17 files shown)


backend/src/database/entities/bcws/bcws-personnel-roles.entity.ts
```diff
@@ -28,11 +28,18 @@ export class BcwsSectionsAndRolesEntity {
   })
   expLevel: ExperienceLevel;
 
+  @Column({
+    name: 'rank',
+    type: 'int',
+  })
+  rank: number;
```
chelsea-EYDS, line 35: I think we should make this one nullable since we're probably only collecting 1st/2nd choice rank (and the num of roles a person could have interest/exp is not limited)

backend/src/database/entities/bcws/bcws-personnel-roles.entity.ts
```diff
@@ -28,11 +28,18 @@ export class BcwsSectionsAndRolesEntity {
...
+
   toResponseObject(): BcwsPersonnelRoleRO {
     return {
       role: this.role.name,
       section: this.role.section,
       expLevel: this.expLevel,
+      rank: this.rank,
```
chelsea-EYDS, line 42: maybe conditional on rank?

backend/src/database/migrations/1715585079303-migration.ts
```diff
@@ -0,0 +1,22 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Migration1715585079303 implements MigrationInterface {
+    name = 'Migration1715585079303'
```
chelsea-EYDS, line 4: name migration? (just to stick with the convention :))

backend/src/personnel/dto/bcws/get-bcws-personnel.dto.ts
```diff
@@ -0,0 +1,60 @@
...
+  )
+  location: string;
+
+  @ApiPropertyOptional({
+    description: 'Section to search personnel from',
+    type: Section,
+    example: Section.COMMAND,
+  })
+  @IsOptional()
+  section: Section;
```
chelsea-EYDS, line 50: I was thinking, do we need to bother including section in the filters as it is implied by the role? (I guess we could potentially have a case where we'd search by section, rather than role + section, and it's optional so probably good to leave it just wanted to check)

backend/src/personnel/personnel.controller.ts
```diff
@@ -104,29 +106,67 @@ export class PersonnelController {
...
+        [Status.INACTIVE]: queryResponse.count[Status.INACTIVE],
+        [Status.PENDING]: queryResponse.count[Status.PENDING],
+      },
+      rows: query.rows,
+      page: query.page,
+    };
+  }
+
+  @ApiOperation({
+    summary: 'Get emcr personnel',
```
chelsea-EYDS, line 148: update summary?

backend/src/personnel/personnel.service.ts
```diff
@@ -192,67 +186,181 @@ export class PersonnelService {
...
+    if (query.location?.length) {
+      qb.andWhere(
+        'bcws_personnel.homeFireCentre.locationName IN (:...homeLocations)',
+        {
+          homeLocations: query.location,
+        },
+      );
+    }
+
+    if (query.section && query.role) {
```
chelsea-EYDS, line 239: I think we can search by role only, as the section is implied by the role, but also totally good to leave it - I wasn't sure if we wanted to pass section along in the query params, or just role.

chelsea-EYDS (commented):
Looks really good just a couple of q's. 

Also, I think this is branched from TC-544? I'm happy to rebase that branch if you would like to point this one to main though? Just lmk 😁

PR #241, opened 2024-05-13
[TC-554] bcws getter

fw-noel:
[TC-554(https://bcdevex.atlassian.net/browse/TC-554)

Objective:
Get endpoint for bcws


backend/src/database/entities/bcws/bcws-personnel-roles.entity.ts
```diff
@@ -28,11 +28,19 @@ export class BcwsSectionsAndRolesEntity {
...
+    nullable: true,
+  })
+  rank: number;
+
   toResponseObject(): BcwsPersonnelRoleRO {
     return {
       role: this.role.name,
       section: this.role.section,
       expLevel: this.expLevel,
+      rank: this.rank,
```
chelsea-EYDS, line 43: if rank is null will this error?

backend/src/personnel/ro/bcws/bcws-personnel-roles.ro.ts
```diff
@@ -31,4 +31,12 @@ export class BcwsPersonnelRoleRO {
...
   })
   expLevel: ExperienceLevel;
+
+  @ApiProperty({
+    description: 'Rank of choice for role (first, second, etc.)',
+    required: true,
+    example: 1,
+    type: 'number'
+  })
+  rank: number;
```
chelsea-EYDS, line 41: I think this will need to be optional since we'll have null for some of the roles' ranks

backend/src/database/entities/bcws/bcws-personnel-roles.entity.ts
fw-noel: It shouldn't; that only happens when we try to extract a value out of a null object

chelsea-EYDS (dismissed):

chelsea-EYDS (approved):

PR #246, opened 2024-05-16
[TC-525] bcws profile page

fw-noel:
[TC-525](https://bcdevex.atlassian.net/browse/TC-525)

Objective:

[TC-525]: https://bcdevex.atlassian.net/browse/TC-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/src/personnel/dto/bcws/create-bcws-personnel.dto.ts
```diff
@@ -171,4 +171,20 @@ export class CreatePersonnelBcwsDTO {
...
   })
   division: number;
+
+  @ApiProperty({
+    description: 'First Choice Section',
+    required: false,
+    enum: Section,
+    example: Section.AVIATION,
+  })
+  firstChoiceSection: Section;
```
chelsea-EYDS, line 181: I think these should be optional?

chelsea-EYDS: (I know they aren't required but sometimes I get an error when I don't put the '?: string' on the field)

chelsea-EYDS (approved):

backend/src/personnel/dto/bcws/create-bcws-personnel.dto.ts
fw-noel: Will merge, can change in the future if it causes an issue

PR #250, opened 2024-05-23
[TC-526]  Profile Edit Form

chelsea-EYDS:
[TC-0526](https://bcdevex.atlassian.net/browse/TC-0526)

Objective:

- first commit includes backend changes for profile edit
- second commit includes frontend changes for profile edit
- third commit is includes profile UI changes


backend/src/database/migrations/1716420585079-ministry-location-migration.ts
```diff
@@ -0,0 +1,107 @@
...
+            where personnel.id = subquery.personnel_id
+        `)
+        await queryRunner.query(`
+            update personnel 
+            set home_location=subquery.id  
+            from (select l.id, ep2.personnel_id from location l join emcr_personnel ep2 on l.location_name = ep2.home_location ) as subquery
+            where personnel.id = subquery.personnel_id
+        `)
+        await queryRunner.query(`
+            update personnel
```
fw-noel, line 32: I can see why the emcr ones need to be formatted like this, but for bcws, since the location ids use `id` as well instead of location name, these two can be simplified right?

frontend/src/common/enums/drivers-license.enum.ts
```diff
@@ -0,0 +1,21 @@
+export enum DriverLicense {
```
fw-noel, line 1: I created this file elsewhere!

fw-noel (approved):
There's an empty file src/hooks/types.tsx, otherwise good

PR #270, opened 2024-05-31
[TC-551]: Pt 1 - form data to test endpoint

chelsea-EYDS:
[TC-0551](https://bcdevex.atlassian.net/browse/TC-0551)

Objective:
- dev form is connected
- parsing form data into our db


fw-noel (dismissed):
Looks good!

PR #285, opened 2024-06-05
[TC-564] emcr date migration

chelsea-EYDS:
[TC-0564](https://bcdevex.atlassian.net/browse/TC-0564)

Objective:

- set date -7 hours (to PST from UTC) for emcr date approved/date applied


fw-noel:
I'm actually thinking this should just be a query we run rather than a migration

PR #319, opened 2024-06-25
[TC-681] pt. 2 - single endpoint for program field data

chelsea-EYDS:
[TC-0681](https://bcdevex.atlassian.net/browse/TC-0681)

Objective:

- Remove extra RO's for role (we had multiple that were similar and all being used in different places)
- single endpoint for public program field data for frontend filters and CHEFS data


backend/src/app.controller.ts
```diff
@@ -107,18 +76,38 @@ export class AppController {
...
-   * Used by CHEFS form
-   * @returns {Ministry[]} List of ministries
+   * Return lists of values used by frontend and CHEFS form
+   * @param req
+   * @returns
    */
   @Public()
-  @Get('/ministries')
-  getMinistries() {
+  @Get('program-field-data')
```
fw-noel, line 84: We had talked about identifying in the request what fields we want to request. Still want to do this?

backend/src/app.controller.ts
```diff
@@ -60,45 +62,12 @@ export class AppController {
...
-    };
-  }
-
   /**
    * Public endpoint with form ID and form enabled status
    * @returns {Object} Form info
    */
-  @Get('/form-info')
   @Public()
+  @Get('/form-info')
```
fw-noel, line 70: Let's avoid dashes in our endpoints

backend/src/bcws/bcws.service.ts
```diff
@@ -238,4 +242,56 @@ export class BcwsService {
...
+  async getAllRoles(): Promise<BcwsRoleEntity[]> {
+    return this.roleRepository.find();
+  }
+  /**
+   * Returns certifications that are not OFA I, II, or III
+   * Used by CHEFS form
+   * @returns {BcwsCertificationEntity[]} List of certifications
+   *
+   */
+  async getCertificates(): Promise<BcwsCertificationEntity[]> {
```
fw-noel, line 259: Would suggest the OFA stuff as a boolean param; `includeOFA` or something

chelsea-EYDS: oh yeah good call

fw-noel (approved):

fw-noel (approved):

PR #341, opened 2024-11-13
[TC-763] BE: Login flow

chelsea-EYDS:
[TC-0763](https://bcdevex.atlassian.net/browse/TC-0763)

Objective:

- BE Changes for supervisor/member login

### DIFF
--- a/backend/src/auth/auth.controller.ts
+++ b/backend/src/auth/auth.controller.ts
@@ -1,0 +1,51 @@
+import {
+  Controller,
+  UseInterceptors,
+  ClassSerializerInterceptor,
+  Inject,
+  Request,
+  Get,
+  HttpStatus,
+} from '@nestjs/common';
+import { ApiOperation, ApiResponse, ApiTags } from '@nestjs/swagger';
+import { AuthService } from './auth.service';
+import { RequestWithRoles, RolesRequest } from './interface';
+import { AppLogger } from '../logger/logger.service';
+
+@Controller('auth')
+@ApiTags('Auth API')
+@UseInterceptors(ClassSerializerInterceptor)
+export class AuthController {
+  constructor(
+    @Inject(AuthService) private readonly authService: AuthService,
+    private readonly logger: AppLogger,
+  ) {
+    this.logger.setContext(AuthController.name);
+  }
+
+  @ApiOperation({
+    summary: 'Verify the role and sent user data',
+    description: 'Returns the user data with the request',
+  })
+  @ApiResponse({
+    status: HttpStatus.OK,
+  })
+  @Get()
+  async checkUserPermissions(
+    @Request() req: RequestWithRoles,
+  ): Promise<RolesRequest> {
+    this.logger.log(
+      `${req.method}: ${req.url} - ${req.role} - ${req.username}`,
+    );
+    const member = await this.authService.getLoggedInUser(req);
+    const supervisor = await this.authService.verifySupervisor(req);
+    return {
+      role: req?.role,
+      program: req?.program,
+      username: `${req?.username}`,
+      idir: req?.idir,
+      member: member !== null,
+      supervisor: supervisor,
+    };
+  }
+}
--- a/backend/src/auth/auth.controller.ts
+++ b/backend/src/auth/auth.controller.ts
@@ -37,15 +37,17 @@ export class AuthController {
     this.logger.log(
       `${req.method}: ${req.url} - ${req.role} - ${req.username}`,
     );
-    const member = await this.authService.getLoggedInUser(req);
-    const supervisor = await this.authService.verifySupervisor(req);
+
+    const { isMember, isSupervisor } =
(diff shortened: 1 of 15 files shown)


backend/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+    );
+    const member = await this.authService.getLoggedInUser(req);
+    const supervisor = await this.authService.verifySupervisor(req);
+    return {
+      role: req?.role,
+      program: req?.program,
+      username: `${req?.username}`,
+      idir: req?.idir,
+      member: member !== null,
+      supervisor: supervisor,
```
chelsea-EYDS, line 48: Might make sense to use an array for Roles but will require more extensive changes. Using role for setting permissions elsewhere, but on this endpoint I am returning member/supervisor boolean in order to know which routes should be available in the user navigation menu.

backend/src/auth/auth.module.ts
```diff
@@ -1,10 +1,31 @@
...
+import { TokenGuard } from './token.guard';
 import { LoggerModule } from '../logger/logger.module';
+import { PersonnelModule } from '../personnel/personnel.module';
 
 @Module({
-  imports: [LoggerModule],
-  providers: [AuthGuard, RolesGuard],
+  imports: [LoggerModule, PersonnelModule],
+  providers: [
+    AuthService,
```
chelsea-EYDS, line 14: Moved these from main into the module in order to avoid explicitly passing in the reflector.

backend/src/personnel/ro/personnel.ro.ts
```diff
@@ -119,12 +118,13 @@ export class PersonnelRO {
...
   supervisorEmail: string;
 
-  @ApiProperty({
-    description: 'Travel preference',
-    required: true,
-    example: TravelPreference.WILLING_TO_TRAVEL_REGION,
-  })
-  travelPreference: TravelPreference;
+  // TODO - confirm if this should be here - I believe this is on the emcr/bcws data? It was throwing an error
+  // @ApiProperty({
```
chelsea-EYDS, line 122: @fw-noel can you let me know if this should be here? It was throwing an error and it looks like we have this set on  the bcws/emcr side?

fw-noel: Yes, this needs to be here for now. They had wanted to add people's preference for travel preference. Not sure why it's throwing an error

backend/src/auth/auth.service.ts
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+      request['username'] = '';
+    }
+    if (payload.email) {
+      request['idir'] = payload.email;
+    } else {
+      request['idir'] = '';
+    }
+
+    const isMember = await this.personnelService.verifyMember(payload.email);
```
fw-noel, line 37: Can shorten this to a single request, get all personnel with email or supervisor as this email. Less db requests the better IMO.

backend/src/personnel/ro/personnel.ro.ts
chelsea-EYDS: okee adding back now sorry!

backend/src/auth/auth.service.ts
chelsea-EYDS: I agree with this in general, but I think it is more readable this way?

chelsea-EYDS: Otherwise we're querying where email === email or where email === supervisor email and then having to figure out whether it was the email that matched or the supervisor email that matched. I'm just trying to keep the role setting as clear as possible since it's part of the auth guard. But open to discussion!

backend/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+    );
+
+    const { isMember, isSupervisor } =
+      await this.authService.verifyMemberOrSupervisor(req.idir);
+
+    return {
+      role: req?.role,
+      program: req?.program,
+      username: `${req?.username}`,
+      idir: req?.idir,
```
fw-noel, line 48: At least with local, username is returning the full name and idir is returning the email. Worth a check if this will be the case in SSO.

backend/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  @ApiOperation({
+    summary: 'Verify the role and sent user data',
+    description: 'Returns the user data with the request',
+  })
+  @ApiResponse({
+    status: HttpStatus.OK,
+  })
+  @Get()
+  async checkUserPermissions(
```
fw-noel, line 34: I'd rename, at this point, we're querying user information. So getUserInfo? Or something

fw-noel (approved):
Minor comments!

backend/src/auth/auth.controller.ts
chelsea-EYDS: copying to address in subsequent pr

PR #345, opened 2024-11-14
[TC-797]: FE Member/Supervisor Login

chelsea-EYDS:
[TC-0797](https://bcdevex.atlassian.net/browse/TC-0797)

Objective:

FE changes for member/supervisor login flow
Moved unauthorized page into the private routes in order to access the user menu top switch routes
Added public unauthenticated page
Added supervisor dash placeholder page
Added member profile placeholder page
Created a wrapper for role protected routes
Updated user menu to show relevant nav items based on the logged in user
created a new hook for the member profile (we might want to change how we're fetching the member data but I think this will require a BE change. We could fetch all profile data from this single endpoint and then exclude based on the logged in user role but this should be worked on separately)

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -70,9 +70,8 @@
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
-      "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
       "dev": true,
+      "license": "Apache-2.0",
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.3.5",
         "@jridgewell/trace-mapping": "^0.3.24"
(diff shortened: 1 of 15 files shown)


frontend/src/providers/Role.tsx
```diff
@@ -1,77 +1,63 @@
...
+  const { AxiosPrivate } = useAxios();
+  const [loading, setLoading] = useState(true)
+  const [user, setUser] = useState<User>()
+
+  const { keycloak } = useKeycloak()
+
+  const getUserPermissions = async () => {
+
+    try {
+
```
fw-noel, line 30: Spacing a lil strange here

frontend/src/routes/Redirect.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Navigate, Outlet } from "react-router"
+import { Routes } from "."
+import { useKeycloak } from "@react-keycloak/web"
+
+const Redirect = () => {
+  const { role } = useRoleContext()
+  const {keycloak} = useKeycloak()
+  if (!keycloak.authenticated) {
+    return <Navigate to={Routes.Home
+    } />
```
fw-noel, line 12: Some more strange spacing, lack of semi colons, etc. on this. Check linter?

fw-noel (approved):
Check your linter, but to make the root the redirect page instead of what we talked about I think is a VERY good idea

PR #376, opened 2024-12-12
[FIX] Supervisor decline approve button state

chelsea-EYDS:

Objective:

Banners + Buttons


fw-noel (dismissed):
It feels like we shouldn't need the green check svg... but if you don't see an alternative, sure

chelsea-EYDS:
> It feels like we shouldn't need the green check svg... but if you don't see an alternative, sure

I agree - I tried to use heroicons for this but I wasn't able to get it to be as thick to match the designs :/

fw-noel (approved):

PR #380, opened 2024-12-13
[FIX] member was redirecting to wrong place

fw-noel:
Objective:
Member only was redirecting to the wrong place


chelsea-EYDS:
ah! Sorry!

chelsea-EYDS (approved):

PR #382, opened 2024-12-17
[FIX] some errors

fw-noel:
Objective:
Getting rid of some of the `placeholder` is missing errors
Combobox removal of isOpen and other invalid props

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -50,7 +50,7 @@
     "@testing-library/jest-dom": "6.2.1",
     "@testing-library/react": "14.1.2",
     "@testing-library/user-event": "14.5.2",
-    "@types/react": "18.2.48",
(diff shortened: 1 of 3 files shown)


frontend/package.json
```diff
@@ -50,7 +50,7 @@
     "@testing-library/jest-dom": "6.2.1",
     "@testing-library/react": "14.1.2",
     "@testing-library/user-event": "14.5.2",
-    "@types/react": "18.2.48",
+    "@types/react": "^18.2.19",
```
chelsea-EYDS, line 53: I don't think it matters here, but do we want to lock the version?

chelsea-EYDS (approved):
No more "placeholder" errors?! nice!!

chelsea-EYDS (approved):

PR #392, opened 2025-01-03
[TC-834] automated emails + testing svc

chelsea-EYDS:
[TC-0834](https://bcdevex.atlassian.net/browse/TC-0834)

Objective:
- automated notifications
- testing endpoint

### DIFF
--- a/backend/src/recommitment/cron.service.ts
+++ b/backend/src/recommitment/cron.service.ts
@@ -3,7 +3,6 @@ import { Cron } from '@nestjs/schedule';
 import { getWeekOfMonth } from 'date-fns';
 import { RecommitmentService } from './recommitment.service';
 import { AppLogger } from '../logger/logger.service';
-import { MailService } from '../mail/mail.service';
 
 export const RecommitmentCron = {
   EVERY_MONDAY_OF_JAN: '0 0 * 1 1',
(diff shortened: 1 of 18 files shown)


backend/src/recommitment/cron.service.ts
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Handles the cron job logic for Mondays in February based on the week of the month.
+   * @param {number} week - The week of the month.
+   */
+  private async handleFebruaryMondayCron(week: number): Promise<void> {
+    if (week === 2) {
+      this.logger.log('2nd Monday in Feb - sending reminder');
+      await this.recommitmentService.handleSendAutomatedReminders();
+    } else if (week === 3) {
+      this.logger.log('3rd Monday in Feb - sending reminder');
```
fw-noel, line 68: Are we doing this for sure? Isn't the third Monday of Feb always after the recommitment period? (ie. 15th at the very earliest). Or did I miss the discussion that it would always end on 3rd Friday of Feb?

fw-noel (approved):

backend/src/recommitment/cron.service.ts
chelsea-EYDS: yeah I did confirm but also I think you're right. I will bring up at DSM.

fw-noel (approved):

PR #402, opened 2025-01-07
[FIX] revised schedule

chelsea-EYDS:

Objective:
- revision of scheduled emails


backend/src/recommitment/cron.service.ts
```diff
@@ -16,39 +16,90 @@ export class CronService {
...
+  @Cron(process.env.START_DATE ?? '0 6 13 1 *', {
     name: 'initial_recommitment',
     timeZone: 'America/Vancouver',
   })
   async initialRecommitment() {
-    const recurringNotificationsJob = new CronJob('0 0 * * 1', async () => {
-      await this.recommitmentService.handleSendAutomatedReminders();
-    });
+    // Start Recommitment. If not in production, use test email and do not send emails to members
+    process.env.ENV === 'production'
```
fw-noel, line 27: Is `ENV` the one that's production across all deployed environments?

fw-noel (dismissed):

backend/src/recommitment/cron.service.ts
chelsea-EYDS: No that one is "NODE_ENV". This one uses dev/test in the those environments.

chelsea-EYDS:
merging as previously approved

PR #412, opened 2025-01-10
[CHORE] Fix jobs - better logging - no undefined

chelsea-EYDS:

Objective:
- simplify the filtering logic
- add more logs
- update limit/request for oc
- fix env var format 
- update tags for easy querying
- update subject line of emails
- create supervisor no response template
- fix supervisor/member reason in the db
- refetch data after submit from banners

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,7 +38,7 @@ CHES_CLIENT_SECRET=
 
 
 
-TEST_EMAIL="chelsea.brown@gov.bc.ca,steffi.pravasi@gov.bc.ca,noel.feliciano@gov.bc.ca"
-END_RECOMMITMENT_SCHEDULE="0 14 * * *"
-START_RECOMMITMENT_SCHEDULE="0 9 * * *"
-NOTIFICATION_SCHEDULE="0 * * * *"
+TEST_EMAIL=chelsea.brown@gov.bc.ca,steffi.pravasi@gov.bc.ca,noel.feliciano@gov.bc.ca
+END_RECOMMITMENT_SCHEDULE=0 14 * * *
+START_RECOMMITMENT_SCHEDULE=0 9 * * *
+NOTIFICATION_SCHEDULE=0 * * * *
--- a/Makefile
+++ b/Makefile
@@ -320,17 +320,14 @@ send-notification-oc:
 	@oc rsh $(SERVER_POD) ./node_modules/.bin/ts-node -e 'require("./dist/jobs/automated_reminders.js")'
 
 create-start-recommitment-cron:
-	@oc process -f openshift/cron-start-recommitment.yml -p APP_NAME=$(APP_NAME) IMAGE_NAMESPACE=$(TOOLS_NAMESPACE) IMAGE_TAG=$(OS_NAMESPACE_SUFFIX) START_RECOMMITMENT_SCHEDULE=$(START_RECOMMITMENT_SCHEDULE) | oc apply -n $(TARGET_NAMESPACE) -f -
+	@oc process -f openshift/cron-start-recommitment.yml -p APP_NAME=$(APP_NAME) IMAGE_NAMESPACE=$(TOOLS_NAMESPACE) IMAGE_TAG=$(OS_NAMESPACE_SUFFIX) START_RECOMMITMENT_SCHEDULE="$(START_RECOMMITMENT_SCHEDULE)" | oc apply -n $(TARGET_NAMESPACE) -f -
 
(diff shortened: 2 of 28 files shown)


backend/src/jobs/automated_reminders.ts
```diff
@@ -33,20 +33,58 @@ export const handler = async () => {
...
+  const today = datePST(new Date())
+
+  logger.log(`Automated Reminders Job: ${today}`, 'Recommitment');
+
+  if (
+    today < recommitment_cycle.endDate ||
+    today > recommitment_cycle.startDate
+  ) {
+    if (process.env.ENV !== 'production') {
+      const testEmails = process.env.TEST_EMAIL.split(',');
```
fw-noel, line 48: Do we need an || clause here in case there's no TEST EMAIL env?

backend/src/mail/mail.service.ts
```diff
@@ -98,36 +100,35 @@ export class MailService {
...
-              ? record.supervisorReason
-              : record.memberReason,
+              ? `${record?.supervisorReason}`
+              : `${record?.memberReason}`,
           subject:
             templateType === TemplateType.SUPERVISOR
-              ? `${record.personnel.supervisorFirstName} ${record.personnel.supervisorLastName}`
-              : `${record.personnel.firstName} ${record.personnel.lastName}`,
-          supervisor: `${record.personnel.supervisorFirstName} ${record.personnel.supervisorLastName}`,
+              ? `${record?.personnel?.supervisorFirstName} ${record?.personnel?.supervisorLastName}`
```
fw-noel, line 129: Are all these `?` necessary? For example, if this is a `records.map`, why would `record` ever be undefined?

openshift/cron-end-recommitment.yml
```diff
@@ -48,10 +48,10 @@ objects:
                 image: image-registry.openshift-image-registry.svc:5000/${IMAGE_NAMESPACE}/${APP_NAME}-server:${IMAGE_TAG}
                 resources:
                   limits:
-                    cpu: 75m
+                    cpu: 40m
```
fw-noel, line 51: These definitely don't need to be this big

fw-noel (approved):

.env.example
```diff
@@ -38,7 +38,7 @@ CHES_CLIENT_SECRET=
 
 
 
-TEST_EMAIL="chelsea.brown@gov.bc.ca,steffi.pravasi@gov.bc.ca,noel.feliciano@gov.bc.ca"
-END_RECOMMITMENT_SCHEDULE="0 14 * * *"
-START_RECOMMITMENT_SCHEDULE="0 9 * * *"
-NOTIFICATION_SCHEDULE="0 * * * *"
\ No newline at end of file
+TEST_EMAIL=chelsea.brown@gov.bc.ca,steffi.pravasi@gov.bc.ca,noel.feliciano@gov.bc.ca
```
chelsea-EYDS, line 41: env vars have to be without quotation marks on local

Makefile
```diff
@@ -320,17 +320,14 @@ send-notification-oc:
 	@oc rsh $(SERVER_POD) ./node_modules/.bin/ts-node -e 'require("./dist/jobs/automated_reminders.js")'
 
 create-start-recommitment-cron:
-	@oc process -f openshift/cron-start-recommitment.yml -p APP_NAME=$(APP_NAME) IMAGE_NAMESPACE=$(TOOLS_NAMESPACE) IMAGE_TAG=$(OS_NAMESPACE_SUFFIX) START_RECOMMITMENT_SCHEDULE=$(START_RECOMMITMENT_SCHEDULE) | oc apply -n $(TARGET_NAMESPACE) -f -
+	@oc process -f openshift/cron-start-recommitment.yml -p APP_NAME=$(APP_NAME) IMAGE_NAMESPACE=$(TOOLS_NAMESPACE) IMAGE_TAG=$(OS_NAMESPACE_SUFFIX) START_RECOMMITMENT_SCHEDULE="$(START_RECOMMITMENT_SCHEDULE)" | oc apply -n $(TARGET_NAMESPACE) -f -
```
chelsea-EYDS, line 323: wrapped vars with quotation marks

backend/src/jobs/automated_reminders.ts
chelsea-EYDS: Good call - mail service will error if we try to send to undefined so I'll return instead

fw-noel (dismissed):

fw-noel (approved):

PR #425, opened 2025-01-13
[TC-855] coord dashboard - member status guide

chelsea-EYDS:
[TC-0855](https://bcdevex.atlassian.net/browse/TC-0855)

Objective:

- showing member status on dashboard
- member status guide on coord dashboard view

### DIFF
--- a/backend/src/emcr/emcr.service.ts
+++ b/backend/src/emcr/emcr.service.ts
@@ -204,6 +204,11 @@ export class EmcrService {
         query.page,
         query.status,
       );
+      personnel.forEach(person => {
+        if(person.personnel.recommitment){
+        person.personnel.recommitment = person?.personnel?.recommitment?.filter(itm => itm.program === Program.EMCR)
+        }
+      })
     return { personnel, count };
   }
   /**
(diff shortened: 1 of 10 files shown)


backend/src/emcr/emcr.service.ts
```diff
@@ -204,6 +204,11 @@ export class EmcrService {
         query.page,
         query.status,
       );
+      personnel.forEach(person => {
+        if(person.personnel.recommitment){
+        person.personnel.recommitment = person?.personnel?.recommitment?.filter(itm => itm.program === Program.EMCR)
```
fw-noel, line 209: This was necessary?

fw-noel (approved):

backend/src/emcr/emcr.service.ts
chelsea-EYDS: Might not be the best way to handle this, but I thought we might not want the bcws data returned from the emcr endpoint, and vice versa.

chelsea-EYDS: I could make the change to filter the entity in the same way as we do with the personnel entity (class transformer + program) but it would require updates everywhere that we're mapping the recommitment entities into the recommitment RO's. (might tech debt that though as it would be cleaner)

chelsea-EYDS: Going to merge just to get it through testing - I have made a note to look at revising this.

PR #445, opened 2025-01-17
[TC-863] - Profile status per program

chelsea-EYDS:
[TC-0863](https://bcdevex.atlassian.net/browse/TC-0863)

Objective:
- profile status tag showing for each program
- updated a couple of button/text colours
- added the grey border around the profile components

### DIFF
--- a/frontend/src/components/profile/header/ProfileMemberHeader.tsx
+++ b/frontend/src/components/profile/header/ProfileMemberHeader.tsx
@@ -1,17 +1,15 @@
 import { useState } from 'react';
 import { Tab, TabsHeader } from '@material-tailwind/react';
 import type { Personnel } from '@/common';
-import { Role, Tabs, TabNames } from '@/common';
-import { PersonnelStatus } from '@/components';
+import { Tabs, TabNames } from '@/common';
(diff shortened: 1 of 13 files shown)


frontend/src/components/profile/header/ProfileMemberHeader.tsx
```diff
@@ -31,9 +31,20 @@ export const ProfileMemberHeader = ({
             {personnel.firstName} {personnel.lastName}
           </h2>
           {roles?.includes(Role.COORDINATOR) && (
-            <span>
-              <PersonnelStatus status={personnel?.status} />
-            </span>
+            <div className="flex flex-col space-y-2">
```
fw-noel, line 34: I think line 33 needs to be removed (my bad!) and this needs to be flex-row? In the design, active in BOTH only says "ACTIVE" but I'm okay with doing it this way for now

chelsea-EYDS: Oh good catch - and yeah I saw that re: if active in both just show one, but it seemed kinda silly to me lol. I will change it though.

fw-noel (approved):

PR #456, opened 2025-01-19
[TC-928] EMCR update

chelsea-EYDS:
[TC-0928](https://bcdevex.atlassian.net/browse/TC-0928)

Objective:
- update emcr personnel from coordinator dash
- TODO: name in the profile header does not refresh when changed, but all other data on page refreshes. (Also, Not sure why the coordinators can change the persons name but the member cannot ??? )

### DIFF
--- a/backend/src/emcr/emcr.controller.ts
+++ b/backend/src/emcr/emcr.controller.ts
@@ -120,7 +120,7 @@ export class EmcrController {
         req.roles,
       );
     } else if (Object.keys(details).length > 0) {
-      return await this.personnelService.updatePersonnel(id, details);
+      return await this.personnelService.updateEmcrPersonnel(id, details);
     } else {
       return await this.personnelService.getEmcrPersonnelById(req.roles, id);
     }
--- a/backend/src/emcr/emcr.controller.ts
+++ b/backend/src/emcr/emcr.controller.ts
@@ -120,11 +120,7 @@ export class EmcrController {
         req.roles,
       );
     } else if (Object.keys(details).length > 0) {
-      return await this.personnelService.updatePersonnel(
-        id,
-        details,
-        req.roles,
-      );
+      return await this.personnelService.updatePersonnel(id, details);
     } else {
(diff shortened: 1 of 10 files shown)


backend/src/emcr/emcr.controller.ts
```diff
@@ -114,15 +114,15 @@ export class EmcrController {
 
     // For now, these are distinct and will not be updated at the same time
     if (experiences) {
-      return this.personnelService.updatePersonnelExperiences(
+      return await this.personnelService.updatePersonnelExperiences(
```
fw-noel, line 117: Still shouldn't need these, weird if its the only way it works..

backend/src/personnel/personnel.service.ts
```diff
@@ -218,31 +217,30 @@ export class PersonnelService {
...
       Object.keys(personnel.bcws).forEach((key) => {
-        bcws[key] = personnel.bcws[key];
+        person.bcws[key] = personnel.bcws[key];
       });
-      console.log(bcws);
     }
     if (personnel.emcr && person.emcr) {
       Object.keys(personnel.emcr).forEach((key) => {
-        emcr[key] = personnel.emcr[key];
+        person.emcr[key] = personnel.emcr[key];
```
fw-noel, line 228: Was this necessary? Thought it was working beforehand..

backend/src/personnel/personnel.service.ts
```diff
@@ -139,12 +139,11 @@ export class PersonnelService {
...
   async updatePersonnelDatabase(
-    email: string,
+    id: string,
     personnel: Partial<CreatePersonnelDTO>,
   ): Promise<PersonnelEntity> {
-    const person = await this.findOneByEmail(email);
-    const bcws = person.bcws;
-    const emcr = person.emcr;
+    const person = await this.findOne(id);
+    delete person.recommitment;
```
fw-noel, line 146: Don't like this. Would rather just query the personnel with the program joins

fw-noel (approved):

PR #465, opened 2025-01-20
[FIX] batch notifications by ministry using ENV

fw-noel:
Objective:
Assign at env for RECOMMITMENT_MINISTRY, if it exists, then only run the recommitment for those members


backend/src/jobs/start_recommitment.ts
```diff
@@ -55,7 +55,13 @@ import { RecommitmentService } from '../recommitment/recommitment.service';
     } else {
       logger.log('Starting recommitment job', 'Start Recommitment');
 
-      const data = await recommitmentService.handleStartRecommitment();
+      const ministry = process.env.RECOMMITMENT_MINISTRY || undefined;
```
chelsea-EYDS, line 58: if we're passing ministry in the test run we'll need to move this up and pass it in

backend/src/recommitment/recommitment.service.ts
```diff
@@ -290,18 +291,30 @@ export class RecommitmentService {
...
-      relations: ['personnel'],
-    });
+    const recommitmentQb =
+      this.recommitmentRepository.createQueryBuilder('recommitment');
+    recommitmentQb.leftJoinAndSelect('recommitment.personnel', 'personnel');
+    recommitmentQb.leftJoinAndSelect(
+      'recommitment.recommitmentCycle',
+      'recommitmentCycle',
+      'recommitmentCycle = :year',
+      { year: cycle.year },
```
chelsea-EYDS, line 311: 'recommitmentCycle.year = :year',

chelsea-EYDS:
You 'll need to add RECOMMITMENT_MINISTRY to the local docker file to test this locally

chelsea-EYDS (approved):

PR #477, opened 2025-01-23
[TC-781] filter active personnel status

chelsea-EYDS:
[TC-781](https://bcdevex.atlassian.net/browse/TC-781)

Objective:
- filter by recom status/all/new

[TC-781]: https://bcdevex.atlassian.net/browse/TC-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/src/personnel/personnel.service.ts
+++ b/backend/src/personnel/personnel.service.ts
@@ -326,6 +326,9 @@ export class PersonnelService {
       queryBuilder.andWhere('recommitment.status = :recommitment_status', {
         recommitment_status: RecommitmentStatus.SUPERVISOR_APPROVED,
       });
+      queryBuilder.andWhere('recommitment.year = :year', {
+        year: new Date().getFullYear(),
+      });
     }
 
     if (name) {
--- a/backend/src/personnel/personnel.service.ts
+++ b/backend/src/personnel/personnel.service.ts
@@ -311,13 +311,19 @@ export class PersonnelService {
     program?: Program,
   ): Promise<SelectQueryBuilder<T>> {
     if (availableStatus && availableStatus === 'New') {
-      program === Program.BCWS
-        ? queryBuilder.andWhere(
-            `bcws_personnel.dateApproved > current_date - interval '5' day `,
-          )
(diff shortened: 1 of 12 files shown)


chelsea-EYDS:
**update bcws service to pass in program**

backend/src/personnel/personnel.service.ts
```diff
@@ -306,7 +307,27 @@ export class PersonnelService {
     availabilityType?: AvailabilityTypeLabel,
     availabilityFromDate?: string,
     availabilityToDate?: string,
+    availableStatus?: string,
+    program?: Program,
   ): Promise<SelectQueryBuilder<T>> {
+    if (availableStatus && availableStatus === 'New') {
+      program === Program.BCWS
```
fw-noel, line 314: Please make an if else rather than ternary, I think I prefer that

fw-noel: Do we need to account for the possibility that ALL could be passed in here? ie. make it if (program === Program.BCWS), else if (program === Program.EMCR)

backend/src/personnel/personnel.service.ts
```diff
@@ -306,7 +307,27 @@ export class PersonnelService {
...
+        ? queryBuilder.andWhere(
+            `bcws_personnel.dateApproved > current_date - interval '5' day `,
+          )
+        : queryBuilder.andWhere(
+            `emcr_personnel.dateApproved > current_date - interval '5' day`,
+          );
+    }
+    if (availableStatus && availableStatus === 'Recommitted') {
+      queryBuilder.andWhere('recommitment.program = :program', {
```
fw-noel, line 323: Do we need to account for `year`?

backend/src/personnel/personnel.service.ts
chelsea-EYDS: This is coming from the personnel dashboard, which is program specific, so I don't think so

fw-noel (approved):

PR #485, opened 2025-01-26
[CHORE] Saving sent mail list in db

chelsea-EYDS:

*deployed and tested on dev*

Objective:
- created two new entities: one for saving a batch record, and one for the individuals messages. 
- If an email address is filtered it will be logged as not sent in order to manually review/address each case
- After sending the emails we can run the check mail job to ensure the status is complete for all sent mail
- Job also checks for any missing entries in the mail logs who should have been emailed based in the recommitment status


backend/src/mail/mail.service.ts
```diff
@@ -164,14 +195,126 @@ export class MailService {
...
+      if (!res.data.txId) {
+        this.logger.error('No txId returned');
+        this.logger.log(res?.data);
+        return;
+      }
+
+      this.logger.log(res?.data);
+
+      const batch =
+        res.data.txId &&
```
fw-noel, line 280: Seems overkill to do this given the check in line 271

backend/src/mail/mail.service.ts
```diff
@@ -164,14 +195,126 @@ export class MailService {
...
 
     try {
       const res = await this.mailApi.post('/emailMerge', mail);
-      this.logger.log(res.data);
+
+      this.logger.log(
+        `Emails sent for ${templateType}: ${res.data.messages.length}`,
+      );
+
+      if (!res.data.txId) {
```
fw-noel, line 271: Need to do `!res?.data?.txId`?

backend/src/mail/mail.service.ts
```diff
@@ -164,14 +195,126 @@ export class MailService {
...
+          return this.mailRepository.create({
+            email: msg.to[0],
+            msgId: msg.msgId,
+            sent: true,
+            txId: batch.txId,
+            tx: batch,
+          });
+        });
+
+      mails.length > 0 && (await this.mailRepository.save(mails));
```
fw-noel, line 301: By and large, I'd prefer using if statements to && just because they're more self explanatory. 

Also, in this case, mails **could** technically be empty from line 289 if, for example, `batch` is undefined for any reason, so this could crash when `mails.length` is searching for length of undefined

backend/src/mail/mail.service.ts
```diff
@@ -164,14 +195,132 @@ export class MailService {
...
+        const existingEmails = existingMail.map((itm) => itm.email);
+        this.logger.log(
+          `Total existing emails to filter: ${existingEmails?.length}`,
+        );
+
+        const filteredContexts = mail.contexts.filter(
+          (itm) => !existingEmails?.includes(itm.to[0]),
+        );
+
+        if (existingEmails?.length > 0) {
```
fw-noel, line 251: `filteredEmails?.length`

backend/src/mail/mail.service.ts
```diff
@@ -164,14 +195,132 @@ export class MailService {
...
+
+      const batch = await this.mailBatchRepository.save(
+        this.mailBatchRepository.create({
+          txId: res.data.txId,
+          tag: mail.contexts[0].tag,
+          template: templateType,
+        }),
+      );
+
+      if (batch.txId && res.data?.messages.length > 0) {
```
fw-noel, line 293: `res.data?.messages?.length`
?

fw-noel (approved):

PR #491, opened 2025-01-29
[FIX] supervisor email lowercase

fw-noel:
Objective:
Update supervisor email in the updateSupervisorInformation function and in creating


chelsea-EYDS:
Looks good - we also do this on the edit profile and edit member profile forms?

fw-noel:
> Looks good - we also do this on the edit profile and edit member profile forms?

It should be, those call updatePersonnelDatabase

chelsea-EYDS (approved):

PR #498, opened 2025-02-05
[FIX] search debounce

chelsea-EYDS:

Objective:
- slight delay on search  field


fw-noel:
Still an issue unfortunately. I'm noticing that if I have availability and dates selected, and click "clear all", and then I enter a name, it sets the availability + dates to the previous value

chelsea-EYDS:
> There is a REALLY EDGE case I'd like you to investigate to see if its a quick fix.
> 
> Steps:
> 
> 1. Select availability and dates such that you get results back
> 2. Refresh the page, so it loads with those query parameters
> 3. Clear All filters
> 4. Quickly add a name to the search after clearing filters
> 
> I notice that it sets the availability + dates again when i enter the name

Can we create a separate ticket for this since it is unrelated to the search debounce? This one was reported as a bug so would be good to get this fix in.

PR #499, opened 2025-02-05
[CHORE] clean up useSupervisor 

chelsea-EYDS:

Objective:
- refactored the hook for fetching data (mostly just some clean up to make the conditionally rendered data parsing more readable)
- added a DTO for search params 
- returning default case for template type (undefined but shouldn't throw an error when destructing the values)
- fixed pagination if less than 4 pages


backend/src/recommitment/recommitment.service.ts
```diff
@@ -251,7 +251,10 @@ export class RecommitmentService {
         };
 
       default:
-        return null;
+        return {
```
fw-noel, line 254: What does this change affect? I don't really see a consequence in this pr

fw-noel (approved):

backend/src/recommitment/recommitment.service.ts
chelsea-EYDS: If the default case is chosen an error was previously thrown since we're destructing the returned value.

chelsea-EYDS: Mostly this was to clean up the code in the useSUpe hook

PR #509, opened 2025-02-11
[TC-960] recommitment reinitation by coord

fw-noel:
[TC-960](https://bcdevex.atlassian.net/browse/TC-960)

Objective:
Allow coordinators to reinitialize recommitment for inactive member
NEEDS MIGRATION

TODO:
- Email content for supervisor
- Tooltip (?) necessary?

[TC-960]: https://bcdevex.atlassian.net/browse/TC-960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/components/ui/classes.ts
+++ b/frontend/src/components/ui/classes.ts
@@ -6,7 +6,7 @@ export const button = {
     'bg-transparent border border-2 rounded-sm  px-6 h-[34px] border-white text-white text-center block  text-xs text-decoration-none whitespace-nowrap hover:opacity-80 disabled:opacity-30',
 
   primaryButton:
-    'bg-transparent border border-2 rounded-sm  px-6 h-[34px] border-primaryBlue text-center block  text-xs text-decoration-none whitespace-nowrap hover:opacity-50 text-primaryBlue hover:text-primaryBlue disabled:opacity-30',
+    'bg-transparent border border-2 rounded-sm  px-6 h-[34px] border-primaryBlue text-center block font-semibold text-xs text-decoration-none whitespace-nowrap hover:opacity-50 text-primaryBlue hover:text-primaryBlue disabled:opacity-30',
 
   tertiaryButton:
     'bg-primaryBlue text-white  rounded-sm  px-6 h-[34px] border-primaryBlue text-center block  text-xs text-decoration-none whitespace-nowrap hover:bg-primaryBlue hover:opacity-80 hover:text-white disabled:bg-[#7E93B3]',
--- a/frontend/src/hooks/usePersonnel.tsx
+++ b/frontend/src/hooks/usePersonnel.tsx
@@ -11,7 +11,7 @@ import { useParams } from 'react-router';
 const usePersonnel = (): {
   personnel: Personnel | undefined;
   updatePersonnel: (person: FormikValues | Personnel) => Promise<void>;
-  fetch: () => void;
+  fetch: () => Promise<void>;
   profileData: ProfileData;
   roles?: Role[];
   loading?: boolean;
--- a/frontend/src/hooks/usePersonnel.tsx
+++ b/frontend/src/hooks/usePersonnel.tsx
@@ -11,6 +11,7 @@ import { useParams } from 'react-router';
 const usePersonnel = (): {
   personnel: Personnel | undefined;
   updatePersonnel: (person: FormikValues | Personnel) => Promise<void>;
+  fetch: () => void;
   profileData: ProfileData;
   roles?: Role[];
   loading?: boolean;
@@ -19,7 +20,7 @@ const usePersonnel = (): {
   const { roles, program, loading } = useRoleContext();
   const [personnel, setPersonnel] = useState<Personnel>();
   const { AxiosPrivate } = useAxios();
-  const [refetch, setRefetch] = useState(false)
+  const [refetch, setRefetch] = useState(false);
   const { profileId } = useParams();
 
   useEffect(() => {
@@ -28,11 +29,16 @@ const usePersonnel = (): {
         const response =
           program && (await AxiosPrivate.get(`/${program}/${profileId}`));
         response && setPersonnel(response.data);
+        setRefetch(false);
       } catch (e) {
         console.log(e);
       }
     })();
   }, [refetch]);
+
+  const fetch = () => {
+    setRefetch(true);
+  };
 
   const updatePersonnel = async (personnel: FormikValues) => {
     if (personnel?.newRoles) {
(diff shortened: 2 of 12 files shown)


frontend/src/components/ui/classes.ts
```diff
@@ -6,7 +6,7 @@ export const button = {
     'bg-transparent border border-2 rounded-sm  px-6 h-[34px] border-white text-white text-center block  text-xs text-decoration-none whitespace-nowrap hover:opacity-80 disabled:opacity-30',
 
   primaryButton:
-    'bg-transparent border border-2 rounded-sm  px-6 h-[34px] border-primaryBlue text-center block  text-xs text-decoration-none whitespace-nowrap hover:opacity-50 text-primaryBlue hover:text-primaryBlue disabled:opacity-30',
+    'bg-transparent border border-2 rounded-sm  px-6 h-[34px] border-primaryBlue text-center block font-semibold text-xs text-decoration-none whitespace-nowrap hover:opacity-50 text-primaryBlue hover:text-primaryBlue disabled:opacity-30',
```
chelsea-EYDS, line 9: Does the BC Gov Sans have a semi bold option?

frontend/src/hooks/usePersonnel.tsx
```diff
@@ -28,12 +29,17 @@ const usePersonnel = (): {
         const response =
           program && (await AxiosPrivate.get(`/${program}/${profileId}`));
         response && setPersonnel(response.data);
+        setRefetch(false);
```
chelsea-EYDS, line 32: If we set refetch to false, wouldn't this trigger another fetch? And then another, and then....

chelsea-EYDS: Oh I guess not, since fetch is already false. Since fetch is false, I am not sure i see the purpose in setting it to false here?

frontend/src/hooks/usePersonnel.tsx
```diff
@@ -28,12 +29,17 @@ const usePersonnel = (): {
...
       } catch (e) {
         console.log(e);
       }
     })();
   }, [refetch]);
 
+  const fetch = () => {
+    setRefetch(true);
```
chelsea-EYDS, line 40: If we call !refetch instead of tracking whether fetch is currently true/false then we can reliably trigger a re-fetch each time that we call setRefetch(!refetch). I'm not sure I understand the intention with explicitly setting true/false (but it is very possible that I am missing something!)

frontend/src/hooks/useRecommitment.tsx
```diff
@@ -55,6 +60,11 @@ export const useRecommitmentCycle = () => {
       recommitmentCycle &&
       offsetTimezoneDate(recommitmentCycle.endDate) >= new Date() &&
       offsetTimezoneDate(recommitmentCycle.startDate) <= new Date(),
+    isRecommitmentReinitiationOpen:
+      recommitmentCycle &&
+      recommitmentCycle.reinitiationEndDate &&
```
chelsea-EYDS, line 65: Do we need to check existence of the re-initiation end date here? I don't think it's hurting anything by including but I was just curious :)

chelsea-EYDS: Oh oops sorry I just looked at the code - in the diff it looks like this is included in the. 'isRecommitmentCycleOpen' check. Resolving :)

frontend/src/pages/profile/Profile.tsx
```diff
@@ -10,24 +10,87 @@ import {
...
+    updateRecommitment,
+  } = useRecommitmentCycle();
+  const [confirmReinitiateOpen, setConfirmReinitiateOpen] = useState(false);
 
   if (loading) {
     return <Loading />;
   }
 
+  const recommitmentStatusText = () => {
+    const recommitment = personnel?.recommitment?.find(
```
chelsea-EYDS, line 42: Looks good - tiny nit - I was wondering if we might consider moving these into the useRecommitment hook? Just for separation of concerns?

chelsea-EYDS (commented):
Looks really good just some minor nits and a couple of place where I made comments and then resolved them :P

frontend/src/hooks/usePersonnel.tsx
chelsea-EYDS: Okay I see why - because we're calling setRefetch(true) below. 

I would feel better about removing the setRefetch(false) and instead in the fetch() function calling setRefetch(!refetch). Reason being, we're calling this elsewhere, so it is entirely possible that the useEffect will be called when setRefetch is true, and would then trigger re-render and re-fetch. I just don't think we can rely on refetch always being false here?

frontend/src/hooks/usePersonnel.tsx
chelsea-EYDS: (sorry this is a duplicate comment to the one above)

frontend/src/components/ui/classes.ts
fw-noel: It seems to work ok

frontend/src/hooks/usePersonnel.tsx
fw-noel: That's smarter actually yeah. I'll add that!

frontend/src/pages/profile/Profile.tsx
fw-noel: Good idea

frontend/src/components/ui/classes.ts
chelsea-EYDS: nice

frontend/src/pages/profile/Profile.tsx
```diff
@@ -10,24 +10,61 @@ import {
...
+        : {
+            emcr: {
+              program: Program.EMCR,
+              year: new Date().getFullYear(),
+              status: RecommitmentStatus.PENDING,
+              reason: '',
+            },
+          };
+    await updateRecommitment(personnel!.id, decision);
+    fetch();
```
chelsea-EYDS, line 64: I think we'll need await fetch

frontend/src/hooks/usePersonnel.tsx
```diff
@@ -11,6 +11,7 @@ import { useParams } from 'react-router';
 const usePersonnel = (): {
   personnel: Personnel | undefined;
   updatePersonnel: (person: FormikValues | Personnel) => Promise<void>;
+  fetch: () => void;
```
chelsea-EYDS, line 14: Promise<void>

chelsea-EYDS (approved):

PR #515, opened 2025-02-13
[FIX] Update deps

chelsea-EYDS:

Objective:
- this should resolve some of the security alerts


backend/package.json
```diff
@@ -38,7 +38,7 @@
     "@nestjs/config": "3.1.1",
     "@nestjs/core": "10.4.7",
     "@nestjs/jwt": "10.2.0",
-    "@nestjs/platform-express": "10.4.7",
+    "@nestjs/platform-express": "^10.4.15",
```
fw-noel, line 41: Pin review

fw-noel (commented):
Please pin versions

backend/package.json
chelsea-EYDS: Updated, but intentionally left the dev dependencies as there is an error on install otherwise (and these won't be installed in production)

PR #519, opened 2025-02-19
[TC-962] reactivation email template

fw-noel:
[TC-962](https://bcdevex.atlassian.net/browse/TC-962)

Email template for reactivating member
Timing changes for using end date vs. reinitiation end date
Use of PST date for the above
Changing other templates to "CORE Team"
Differentiate between member steps and member follow up steps as Chevonne mentioned they are intentionally different

[TC-962]: https://bcdevex.atlassian.net/browse/TC-962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/src/recommitment/recommitment.service.ts
+++ b/backend/src/recommitment/recommitment.service.ts
@@ -287,7 +287,7 @@ export class RecommitmentService {
       case RecommitmentStatus.PENDING:
         return {
           templateType: EmailTags.MEMBER_REACTIVATE,
-          templateRecipient: TemplateType.SUPERVISOR,
+          templateRecipient: TemplateType.MEMBER,
         };
       case RecommitmentStatus.MEMBER_COMMITTED:
         return {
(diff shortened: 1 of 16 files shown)


backend/src/recommitment/recommitment.service.ts
```diff
@@ -270,6 +284,11 @@ export class RecommitmentService {
     templateRecipient: TemplateType;
   } {
     switch (status) {
+      case RecommitmentStatus.PENDING:
+        return {
```
chelsea-EYDS, line 288: Just want to double check that the member recommitment status is updated as PENDING in recommitment table prior to calling this?

chelsea-EYDS (dismissed):

chelsea-EYDS (approved):

PR #520, opened 2025-02-19
[FIX] Update section for member of single program

chelsea-EYDS:

Objective:
- split out the FE components for member update roles/function preferences
- fix update for single member
- for EMCR add the function to the functions array (required for filtering)


backend/src/personnel/personnel.service.ts
```diff
@@ -185,20 +185,20 @@ export class PersonnelService {
...
+        emcr: preferences.emcr?.experiences
           ? {
               experiences: true,
             }
           : false,
       },
     });
 
-    if (preferences.bcws) {
+    if (preferences?.bcws) {
```
fw-noel, line 201: It's fine, but preference seems to be non nullable as the function's parameter so this shouldn't be necessary

fw-noel (approved):

backend/src/personnel/personnel.service.ts
chelsea-EYDS: You're right - I got a little carried away, lol

fw-noel (approved):

PR #522, opened 2025-02-19
[TC-949] Inactive Filter Recommitment

leon-ey:
Ticket:
https://emcr.atlassian.net/browse/TC-949

Dependencies:
Backend (add here later)

Description:
- Set up the filter selector for the `INACTIVE` tab on the dashboard, similar to the `ACTIVE` tab
- Refactored the code to use `RecommitmentStatusFilter` values instead of `StatusFilter` values
- Made `StatusFilter` more reusable by providing `statusFilter` prop (takes a `[{ value: string, label: string }]` object
- Added `searchParams.set(Filters.AVAILABLE_STATUS, 'ALL');` to `changetab` on `Tab` so that the selector flips to `All` whenever the tab is changed

Notes:
- No backend changes here, so the filters will not work (currently, the backend is expecting different values, for example `All` instead of `ALL`, we changed it to `RecommitmentFilter` values in here)

Screenshots:
image
image


backend/src/personnel/personnel.service.ts
```diff
@@ -346,10 +346,11 @@ export class PersonnelService {
     availabilityType?: AvailabilityTypeLabel,
     availabilityFromDate?: string,
     availabilityToDate?: string,
-    availableStatus?: string,
+    availableStatus?: AvailabilityTypeStatus,
     program?: Program,
   ): Promise<SelectQueryBuilder<T>> {
-    if (availableStatus && availableStatus === 'New') {
+    console.log(availableStatus, 'STATUS');
```
chelsea-EYDS, line 352: console (my bad, lol)

frontend/src/common/enums/recommitment-status.ts
```diff
@@ -61,3 +61,16 @@ export enum RecommitmentStatusLabel {
...
 }
+
+export const RecommitmentStatusFilter = [
+  { value: 'ALL', label: 'All' },
+  { value: 'NEW', label: 'New' },
+  { value: 'ACTIVE', label: 'Active' },
+  { value: 'INACTIVE', label: 'Inactive' },
+  { value: 'MEMBER_COMMITTED', label: 'Recommitted' },
+  { value: 'MEMBER_DENIED', label: 'Member Declined - Not Returning' },
+  { value: 'MEMBER_NO_RESPONSE', label: 'Missed Recommitment' },
```
chelsea-EYDS, line 72: We're going to need one more change here - "Missed Recommitment" needs to cover two cases: member no response and supervisor no response

frontend/src/common/enums/status.enum.ts
```diff
@@ -7,9 +7,26 @@ export enum Status {
 
 export enum StatusNames {
   ACTIVE = 'Active',
+  ALL = 'All',
   INACTIVE = 'Inactive',
-  PENDING = 'Pending',
+  MEMBER_DECLINED = 'MemberDeclined',
+  MISSED = 'Missed',
```
chelsea-EYDS, line 13: I don't think we're using these anymore

chelsea-EYDS (approved):

fw-noel (approved):

PR #525, opened 2025-02-20
[TC-949] Feat/tc 949 inactive filter recommitment pt 2

chelsea-EYDS:
[TC-0949](https://bcdevex.atlassian.net/browse/TC-0949)

Objective:
- "Recommitted" filter value now returning approved members
- "Missed Deadline" filter value now returning members who missed their own deadline + members whose supervisors missed the deadline


frontend/src/pages/dashboard/Dashboard.tsx
```diff
@@ -97,28 +97,9 @@ const Dashboard = () => {
...
-                statusFilter={RecommitmentStatusFilter.filter((obj) =>
-                  [
-                    'ALL',
-                    'MEMBER_NO_RESPONSE',
-                    'MEMBER_DENIED',
-                    'SUPERVISOR_DENIED',
-                    'OTHER',
-                  ].includes(obj.value),
-                )}
+                statusFilter={searchParams.get(Filters.STATUS) && searchParams.get(Filters.STATUS) === Status.ACTIVE ? ActiveRecommitmentStatusFilter : InactiveRecommitmentStatusFilter}
```
chelsea-EYDS, line 102: @leon-ey  the way you had this previously was fine but whenever we can it is preferable to reduce inline logic so I have changed this to use two separately defined constants.

frontend/src/pages/dashboard/Dashboard.tsx
```diff
@@ -77,7 +77,7 @@ const Dashboard = () => {
                 <Tabs
                   tabs={tabs}
                   changeTab={(value: unknown) => {
-                    searchParams.set(Filters.AVAILABLE_STATUS, 'ALL');
+                    searchParams.delete(Filters.AVAILABLE_STATUS);
```
chelsea-EYDS, line 80: Deleting instead of setting - same outcome, fewer params passed into the api request

leon-ey:
Makes sense to me, checked the filters again, feel free to merge, thank you for your work!

PR #528, opened 2025-02-20
[FIX] BE changes for supervisor/ member decision

chelsea-EYDS:

Objective:

- split the recommitment end points


backend/src/recommitment/recommitment.service.ts
```diff
@@ -39,7 +38,118 @@ export class RecommitmentService {
...
+        : recommitmentCycle.endDate;
+
+    const recommitment = await this.recommitmentRepository.findOneByOrFail({
+      personnelId: id,
+      recommitmentCycleId: new Date().getFullYear(),
+      program: key,
+    });
+
+    const emailTemplate = this.mailService.generateTemplate(
+      EmailTags.MEMBER_REACTIVATE,
```
fw-noel, line 87: Don't love this. Makes the assumption of the status passed in. Same with the corresponding segment in the supervisor function.

I would add an if statement at the start of each function (or the controller endpoint), which throws an error if the update is not one of the expected statuses.

fw-noel (approved):

PR #530, opened 2025-02-21
[FIX] end recommitment update

chelsea-EYDS:

Objective:
- end recommitment


backend/src/emcr/emcr.service.ts
```diff
@@ -42,14 +42,14 @@ export class EmcrService {
    * @param status
    */
   async updatePersonnelAfterRecommitment(
-    id: string,
+    id: string[],
```
fw-noel, line 45:
```suggestion
    ids: string[],
```

fw-noel (dismissed):

chelsea-EYDS:
prev approved

PR #531, opened 2025-02-21
[FIX] fix env var

chelsea-EYDS:

Objective:
- FIX env var and cron yml


chelsea-EYDS:
verbal approval from noel

PR #532, opened 2025-02-21
[TC-925] fix: add default placeholder for pending

leon-ey:
Ticket:
https://emcr.atlassian.net/browse/TC-925

Description:
- Added a default empty `MembershipDetails` object that applies to the personnel who don't have entries in the `recommitment` table
- This makes sure that the placeholder empty data is still displayed, otherwise the whole section is empty and it was confusing for the client

Screenshots:
1) Before:
image

2) After:
image

3) With data present in `recommitment` for that `personnel_id`:
image

### DIFF
--- a/frontend/src/hooks/commonProfileData.tsx
+++ b/frontend/src/hooks/commonProfileData.tsx
@@ -1,8 +1,17 @@
-import { DriverLicenseName, DriverLicense, Member, Program, Personnel } from "@/common";
-import { LanguageProficiencyName } from "@/common/enums/language.enum";
-import { RecommitmentStatus } from "@/common/enums/recommitment-status";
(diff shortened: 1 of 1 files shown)


frontend/src/hooks/commonProfileData.tsx
```diff
@@ -59,6 +93,51 @@ export const renderRecommitmentStatus = (
...
+        title: 'Status',
+        content: (status && renderStatus(status)) || '--',
+      },
+      {
+        title: 'Annual Recommitment',
+        content: '--',
+      },
+      {
+        title: 'Member since',
+        content: '--',
```
chelsea-EYDS, line 115: and this as: `personnel.emcr.dateApproved` ?

chelsea-EYDS (approved):

PR #539, opened 2025-02-26
[TC-1085] reset recommitment for both programs

fw-noel:
[TC-1085](https://bcdevex.atlassian.net/browse/TC-1085)

Objective:

### DIFF
--- a/backend/src/mail/constants.ts
+++ b/backend/src/mail/constants.ts
@@ -38,7 +38,7 @@ export enum EmailSubjects {
   MEMBER_FOLLOW_UP = 'ACTION REQUIRED: Confirm or Decline Recommitment for the {{year}} CORE Program',
   MEMBER_NO_RESPONSE = 'CORE TEAM: Notice for Change in Member Status',
   MEMBER_ANNUAL = 'ACTION REQUIRED: Annual Recommitment Reminder. Confirm or Decline Recommitment for the {{year}} CORE Program',
-  MEMBER_REACTIVATE = 'ACTION REQUIRED: Update your CORE Recommitment status for {{year}}',
+  MEMBER_REACTIVATE = 'ACTION REQUIRED: Confirm or Decline Recommitment for the {{year}} CORE Program',
   MEMBER_SUPERVISOR_NO_RESPONSE = 'CORE TEAM: Notice for Member Status Change',
   SUPERVISOR_ANNUAL = 'CORE Team {{year}} Annual Recommitment Period Has Started',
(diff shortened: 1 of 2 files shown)


backend/src/mail/constants.ts
```diff
@@ -38,7 +38,7 @@ export enum EmailSubjects {
   MEMBER_FOLLOW_UP = 'ACTION REQUIRED: Confirm or Decline Recommitment for the {{year}} CORE Program',
   MEMBER_NO_RESPONSE = 'CORE TEAM: Notice for Change in Member Status',
   MEMBER_ANNUAL = 'ACTION REQUIRED: Annual Recommitment Reminder. Confirm or Decline Recommitment for the {{year}} CORE Program',
-  MEMBER_REACTIVATE = 'ACTION REQUIRED: Update your CORE Recommitment status for {{year}}',
+  MEMBER_REACTIVATE = MEMBER_FOLLOW_UP,
```
chelsea-EYDS, line 41: Hey I just wanted to make sure you tested this send? I was getting and error previously when using the same value here (which I assume is why you're setting it to the enum, clever, just wanted to double check :) )

backend/src/recommitment/recommitment.service.ts
```diff
@@ -84,12 +84,43 @@ export class RecommitmentService {
...
+      await this.emcrService.getEmcrPersonnelById([Role.COORDINATOR], id);
+    let otherProgramStatusReset = false;
+    if (otherProgramPersonnel) {
+      const recommitmentOtherProgram = await this.recommitmentRepository.findOneByOrFail({
+        personnelId: id,
+        recommitmentCycleId: new Date().getFullYear(),
+        program: otherProgram,
+      });
+      // Unless they were already approved for the other program, reset to PENDING
+      if (recommitmentOtherProgram.status !== RecommitmentStatus.SUPERVISOR_APPROVED) {
```
chelsea-EYDS, line 99: Are we still showing the recommitment reinit button on the FE if a user becomes INACTIVE for other reasons, but had been prev. approved by their supervisor?

chelsea-EYDS (dismissed):

backend/src/recommitment/recommitment.service.ts
fw-noel: I think so, yes. We leave it up to the Coordinator which pathway they want to go through for the member's reactivation

backend/src/mail/constants.ts
fw-noel: Yep, it seems to :)

chelsea-EYDS (approved):

PR #552, opened 2025-03-06
[TC-949] fix: change availabilityStatus filters

leon-ey:
Ticket:
https://emcr.atlassian.net/browse/TC-949

Description:
- Changed the inactive filters on the frontend in accordance to the ticket's specifications
- Fixed the backend to reflect the frontend changes
- Fixed an unrelated date formatting bug that would crash the member profile page


backend/src/personnel/personnel.service.ts
```diff
@@ -390,15 +390,20 @@ export class PersonnelService {
       });
     } else if (
```
chelsea-EYDS, line 391: This comment is not related to you PR but can you add an "else" on line 378?
Screenshot 2025-03-06 at 12 57 00 PM

leon-ey: Sure thing, added the missing "else" statement

chelsea-EYDS (approved):

PR #553, opened 2025-03-06
[TC-1072] chips endpoint

fw-noel:
[TC-1072](https://bcdevex.atlassian.net/browse/TC-1072)

Objective:
CHIPS endpoint
Will only work on Dev / Test and with a subset of people (ie. none of the EY IDIRs will work with it)


backend/src/app.controller.ts
```diff
@@ -117,15 +117,24 @@ export class AppController {
     return await this.recommitmentService.checkRecommitmentPeriod();
   }
 
+  @Public()
```
chelsea-EYDS, line 120: Should this be public?

backend/src/app.controller.ts
```diff
@@ -117,15 +117,24 @@ export class AppController {
...
+  @ApiQuery({
+    name: 'email',
+    type: String,
+    description: 'Email',
+    required: false
+  })
+  async chips(
+    @Query("email") email?: string,
+  ) {
     if (process.env.ENV !== 'dev') {
```
chelsea-EYDS, line 131: Is this endpoint only for dev? No test or prod?

backend/src/database/entities/personnel/personnel.entity.ts
```diff
@@ -212,6 +212,57 @@ export class PersonnelEntity extends BaseEntity {
...
+    default: false,
+  })
+  chipsProfileMissing: boolean;
+
+  @Column({
+    name: 'chips_last_updated_properties',
+    type: 'jsonb',
+    default: () => "'{}'",
+  })
+  chipsLastUpdatedProperties: { [key: string]: string | object };
```
chelsea-EYDS, line 241: Should we define interfaces with the keys we will be using in the json? Just for readability

backend/src/database/migrations/1741166137543-chips-migration.ts
```diff
@@ -0,0 +1,26 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class ChipsMigration1741166137543 implements MigrationInterface {
+    name = 'ChipsMigration1741166137543'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "personnel" ADD "chips_last_ping" date`);
+        await queryRunner.query(`ALTER TABLE "personnel" ADD "chips_last_action_date" date`);
+        await queryRunner.query(`ALTER TABLE "personnel" ADD "chips_profile_missing" boolean NOT NULL DEFAULT false`);
```
chelsea-EYDS, line 9: Just double checking these be not null? What if someone has no chips profile? (you can disregard and mark as resolved if this is correct -just double checking)

backend/src/personnel/personnel.service.ts
```diff
@@ -992,4 +998,112 @@ export class PersonnelService {
 
     return { emcr: activeEmcrPersonnel, bcws: activeBcwsPersonnel };
   }
+
+  async updatePersonnelChipsData(personnel: PersonnelEntity, data: ChipsResponse) {
```
chelsea-EYDS, line 1002: Can/should we have a separate CHIPS service?

chelsea-EYDS:
import region/location module into personnel module should fix the tests

backend/src/app.controller.ts
fw-noel: Definitely not prod, this is really just for testing the chips endpoint. It shouldn't really be in anything at all eventually.

backend/src/database/migrations/1741166137543-chips-migration.ts
fw-noel: Not null is correct when profile missing defaults to false and will turn true
Not null for the others is fine I think, since I'm setting a default value. They're pretty safely ignored when there's no chips profile

docker-compose.yml
```diff
@@ -98,6 +98,7 @@ services:
       TEST_RUN: ${TEST_RUN}
       VIEWS: ${VIEWS}
       RECOMMITMENT_MINISTRY: ${RECOMMITMENT_MINISTRY}
+      TEST_CHIPS_RESPONSE: ${TEST_CHIPS_RESPONSE}
```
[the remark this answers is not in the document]
fw-noel: Oh yeah good call

backend/src/database/entities/personnel/personnel.entity.ts
chelsea-EYDS: Hmmm... could we define an interface with optional values? Reason being, if we need to query the jsonb column later, it helps to know  key names. Otherwise, maybe a comment here to explain this could be any key of personnel?

chelsea-EYDS (dismissed):

chelsea-EYDS (approved):

PR #558, opened 2025-03-10
[TC-1106] chips job

fw-noel:
[TC-1106](https://bcdevex.atlassian.net/browse/TC-1106)

Objective:

### DIFF
--- a/backend/src/jobs/chips.ts
+++ b/backend/src/jobs/chips.ts
@@ -8,35 +8,29 @@ import { PersonnelService } from '../personnel/personnel.service';
 import { PersonnelEntity } from '../database/entities/personnel/personnel.entity';
 import { Status } from '../common/enums';
 import { isAfter } from 'date-fns';
+import { ChipsAppModule } from '../modules/chips-app.module';
 
 
 (async () => {
   if (!datasource) {
     await datasource.initialize();
   }
-  const app = await NestFactory.create<NestExpressApplication>(AppModule, {
+  const app = await NestFactory.create<NestExpressApplication>(ChipsAppModule, {
     rawBody: true,
     bufferLogs: true,
   });
 
   const logger = new AppLogger();
   try {
-    const personnelRepository = datasource.getRepository(PersonnelEntity);
-    const qb = personnelRepository.createQueryBuilder('personnel');
-    qb.leftJoinAndSelect('personnel.emcr', 'emcr');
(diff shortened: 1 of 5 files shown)


backend/src/jobs/chips.ts
```diff
@@ -0,0 +1,59 @@
...
+import { PersonnelEntity } from '../database/entities/personnel/personnel.entity';
+import { Status } from '../common/enums';
+import { isAfter } from 'date-fns';
+
+
+(async () => {
+  if (!datasource) {
+    await datasource.initialize();
+  }
+  const app = await NestFactory.create<NestExpressApplication>(AppModule, {
```
chelsea-EYDS, line 17: I think if you're not using nunjucks template here you can use createNestContext, rather than NestFactory.create (I _think_ it's a bit more lightweight? Could be wrong - it is the usual recommended approach for jobs/lambdas etc)

openshift/cron-chips.yml
```diff
@@ -0,0 +1,203 @@
...
+  - name: CONFIG_VERSION
+    description: The github repository version of Config (Commit SHA)
+    value: default
+objects:
+  - apiVersion: batch/v1
+    kind: CronJob
+    metadata:
+      name: ${APP_NAME}-cronjob-chips
+    spec:
+      schedule: 0 0 0 0 0
```
chelsea-EYDS, line 29: `0 0 0 * *`   ?

openshift/cron-chips.yml
```diff
@@ -0,0 +1,203 @@
...
+      concurrencyPolicy: 'Replace'
+      startingDeadlineSeconds: 10
+      successfulJobsHistoryLimit: 1
```
chelsea-EYDS, line 32: Since these pods won't be counted in our long time compute quota we might want to keep 3 for logs?

chelsea-EYDS: (Just since this one is nightly - I'm not sure we retain the logs otherwise)

openshift/cron-chips.yml
fw-noel: Leaving it at this, where we'll be changing the schedule on Openshift after some finalizing

openshift/cron-chips.yml
fw-noel: I'm okay with that

chelsea-EYDS (dismissed):

chelsea-EYDS (approved):

PR #559, opened 2025-03-10
[TC-1003] Form UI Progress

leon-ey:
Work in progress on the fields' UI

Description:
- Added some more fields to the form
- Added checkbox and radio buttons
- Added some extra formatting options to the tab fields
- Added validations for those fields (except for some checkmarks)

Screenshots:
image
image


frontend/src/hooks/useIntakeForm.tsx
```diff
@@ -25,7 +25,7 @@ export const useIntakeForm = () => {
     })();
   }, []);
 
-  const saveUpdateForm = async (values: IntakeFormPersonnelData) => {
+  const saveUpdateForm = async (values: any) => {
```
chelsea-EYDS, line 28: Let's revert this to use the interface

frontend/src/pages/intake-form/FormField.tsx
```diff
@@ -5,10 +5,20 @@ import { renderField } from './helpers';
...
+    options,
+    helper,
+    colspan,
+    hideLabel,
+  } = field;
   return (
-    <div key={name}>
-      <label htmlFor={name}>
+    <div key={name} className={colspan ? `col-span-${colspan}` : ''}>
+      <label htmlFor={name} className={hideLabel ? 'invisible' : ''}>
```
chelsea-EYDS, line 21: We should be able to pass an empty string here rather than setting the class conditionally.

frontend/src/pages/intake-form/components/SelectField.tsx
```diff
@@ -51,7 +51,7 @@ export const SelectField = (
   const options = getOptions();
 
   return (
-    <select className={classes.menu.container} {...props} {...field}>
+    <select className={classes.menu.container} {...props} {...field} value={undefined} defaultValue={''}>
```
chelsea-EYDS, line 54: is value/default value needed here? Setting the value explicitly will change this from uncontrolled to controlled input

frontend/src/pages/intake-form/helpers.tsx
```diff
@@ -19,7 +19,7 @@ export const renderField = (
       return <SelectField {...fieldItm} {...fieldProps} />;
     case 'radio':
       return <RadioGroupField {...fieldItm} {...fieldProps} />;
-    case 'checkbox-group':
+    case 'checkbox':
```
chelsea-EYDS, line 22: Can we keep this as checkbox group? Just in case we need a single checkbox field later

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -224,9 +225,106 @@ export const formTabs: FormTab[] = [
...
+            program: 'bcws',
+          },
+          {
+            name: 'travelPreference',
+            label: 'Travel Preferences (for deployment',
+            type: 'select',
+            required: true,
+            placeholder: 'Select an option',
+            options: [
+              { label: 'Option 1', value: 'travel1' },
```
chelsea-EYDS, line 284: We'll need two separate fields here for travel preference which are program specific. You can populate the values by mapping over Object.values(TravelPreferencesEmcr / travelPreferencesBcws). 

I know the designs show a single field here but in the database this field is on the bcws_personnel/emcr_personnel tables and has a specific enum for each.

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -224,9 +225,106 @@ export const formTabs: FormTab[] = [
...
+            required: false,
+            placeholder: 'Smith',
+            program: 'bcws',
+          },
+          {
+            name: 'liaisonEmail',
+            label: 'Liaison Email',
+            type: 'text',
+            required: false,
+            placeholder: 'johnsmith@gov.bc.ca',
```
chelsea-EYDS, line 266: Can we confirm the placeholder text should be examples and not "Enter Email" etc? (Just double check with Chevonne)

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -224,9 +225,106 @@ export const formTabs: FormTab[] = [
...
+          {
+            name: 'emergencyPhone',
+            label: 'Emergency Contact Phone Number',
+            type: 'tel',
+            required: true,
+            placeholder: '000-000-0000',
+          },
+          {
+            name: 'emergencyRelationship',
+            label: 'Emergency Contract Relationship',
```
chelsea-EYDS, line 316: typo

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -224,9 +225,106 @@ export const formTabs: FormTab[] = [
...
+            name: 'liaisonPhoneNumber',
+            label: 'Liaison Phone Number',
+            type: 'tel',
+            required: false,
+            placeholder: '000-000-0000',
+            program: 'bcws',
+          },
+          {
+            name: 'travelPreference',
+            label: 'Travel Preferences (for deployment',
```
chelsea-EYDS, line 279: label typo?

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -224,9 +225,106 @@ export const formTabs: FormTab[] = [
             required: false,
             placeholder: '000-000-0000',
           },
+          {
+            name: 'liaisonUnknownCheckbox',
+            label: 'I am unsure who my liaison is',
```
chelsea-EYDS, line 230: Set label = ''

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -224,9 +225,106 @@ export const formTabs: FormTab[] = [
...
+            required: true,
+            placeholder: '000-000-0000',
+          },
+          {
+            name: 'emergencyRelationship',
+            label: 'Emergency Contract Relationship',
+            type: 'select',
+            required: true,
+            placeholder: 'Select an option',
+            options: [
```
chelsea-EYDS, line 320: This field should just be a text input.

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -276,9 +423,64 @@ export const formTabs: FormTab[] = [
           {
             name: 'emcr.functions',
             label: 'Emcr Function Experiences',
-            type: 'checkbox-group',
-            required: true,
-            options: [],
+            type: 'checkbox',
+            required: false,
+            placeholder: '',
+            program: 'emcr',
```
chelsea-EYDS, line 429: let's set program using enum: `Program.EMCR`

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -276,9 +423,64 @@ export const formTabs: FormTab[] = [
...
+            colspan: 2,
+            hideLabel: true,
```
chelsea-EYDS, line 431: label: ""

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -276,9 +423,64 @@ export const formTabs: FormTab[] = [
...
+            options: [
+              {
+                label: 'Advance Planning Unit',
+                value: 'false',
```
chelsea-EYDS, line 435: Let's set default=false on the checkbox rather than defining the value in the field?

frontend/src/pages/intake-form/tabs.tsx
```diff
@@ -316,7 +518,7 @@ export const formTabs: FormTab[] = [
           {
             name: 'bcws.roles',
             label: 'BCWS Roles',
-            type: 'checkbox-group',
+            type: 'checkbox',
```
chelsea-EYDS, line 521: 'checkbox-group'

frontend/src/pages/intake-form/validation.ts
```diff
@@ -170,4 +170,26 @@ export const intakeFormValidationSchema = Yup.object().shape({
   supervisorLastName: Yup.string().required("Supervisor's last name is required"),
   supervisorEmail: Yup.string().required("Supervisor's email is required"),
   supervisorPhone: Yup.string(),
+
+  liaisonUnknown: Yup.string(),
+  liaisonFirstName: Yup.string(),
```
chelsea-EYDS, line 175: Custom validation message on these fields is needed. You can also set .optional()

frontend/src/pages/intake-form/FormField.tsx
```diff
@@ -5,10 +5,20 @@ import { renderField } from './helpers';
...
+    placeholder,
+    options,
+    helper,
+    colspan,
+    hideLabel,
+  } = field;
   return (
-    <div key={name}>
-      <label htmlFor={name}>
+    <div key={name} className={colspan ? `col-span-${colspan}` : ''}>
```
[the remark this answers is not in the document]
leon-ey: Moved to the parent component.

frontend/src/pages/intake-form/tabs.tsx
leon-ey: image
That's what it is in Formik, but I can confirm.

frontend/src/pages/intake-form/tabs.tsx
leon-ey: In this case, if we pick both, do we make two selectors: `EMCR - Travel Preferences (for deployment)` and `BWCS - Travel Preferences (for deployment)`?

frontend/src/pages/intake-form/tabs.tsx
leon-ey: image

frontend/src/pages/intake-form/validation.ts
```diff
@@ -75,10 +75,10 @@ export const emergencyContactDetailsSchema = Yup.object().shape({
 });
 
 export const generalEmergencyManagementExperienceSchema = Yup.object().shape({
-  directExperience: Yup.boolean().required(),
+  emergencyExperience: Yup.boolean().required(),
   preocExperience: Yup.boolean().required(),
   peccExperience: Yup.boolean().required(),
-  indigenousExperience: Yup.boolean().required(),
+  firstNationsWorking: Yup.boolean().required(),
```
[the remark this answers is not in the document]
leon-ey: It doesn't seem like we have used `firstNationsLiving` anywhere on front or backend, is it a new field that needs to be added?
image

frontend/src/pages/intake-form/tabs.tsx
leon-ey: value is required for all our options, should we make it optional? Or make a separate `options` type for checkboxes that has `label` and `name` but not `value`?

frontend/src/pages/intake-form/tabs.tsx
chelsea-EYDS: Contract -> Contact (it may have been my typo initially)

frontend/src/pages/intake-form/validation.ts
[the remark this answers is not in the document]
chelsea-EYDS: It's just one of the form fields. We don't use living or working but we're collecting the data.

frontend/src/pages/intake-form/validation.ts
chelsea-EYDS: Anything you think is fine, we just want to show a nice message incase they try to input non string values in the optional fields.

frontend/src/pages/intake-form/validation.ts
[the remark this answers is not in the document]
leon-ey: What would be the question in this case?
`Do you have any direct experience living with Indigenous communities?`
If so, seem like the one above already covers it, as is says "living, working, etc."
image

frontend/src/pages/intake-form/tabs.tsx
leon-ey: Sorry maybe I don't understand it, do you mean putting `Checkbox.FALSE` instead of `false`, where `Checkbox` is an enum of just two values: `true` and `false`? Let me know!

frontend/src/pages/intake-form/tabs.tsx
leon-ey: Oh my bad, great catch, thank you!

frontend/src/pages/intake-form/helpers.tsx
leon-ey: Should we rename radio to radio-group in this case to be consistent?
Also, I thought about it and that's how I set it up initially, but wouldn't a single checkbox be a checkbox-group with just one option? That's how I set it up for the unknownLIaison thing. Let me know what you think!

frontend/src/pages/intake-form/components/SelectField.tsx
leon-ey: This is what happens if you don't set it up like this:
image

It sets it to the first option in the list, and I am pretty sure the clients will complain about it. In their design, it should say "select an option".
Also, if you don't pick anything, it will still look like you selected an option, but it didn't validate:
image


Let me know what you think!

frontend/src/pages/intake-form/helpers.tsx
chelsea-EYDS: Yes, let's rename "radio" to "radio-group" good call

frontend/src/pages/intake-form/tabs.tsx
chelsea-EYDS: Ideally we want to have an array of values for the options which are populated from the backend/DB, and for the form submission we'd be passing in the array of selected values from this list.

frontend/src/pages/intake-form/validation.ts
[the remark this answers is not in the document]
chelsea-EYDS: Let's ask Chevonne re: copy, otherwise the original CHEFS form will have this

frontend/src/pages/intake-form/helpers.tsx
leon-ey: Awesome, thank you! Renamed them accordingly.

frontend/src/pages/intake-form/validation.ts
chelsea-EYDS: Hrm okay as long as there is not weird looking validation msg displayed then we should be g2g

frontend/src/pages/intake-form/validation.ts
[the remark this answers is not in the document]
chelsea-EYDS: nvm just checked the form - this value is no longer on the form, but it is in our database. I will follow up with the team to confirm what we should do with this item in our db.

frontend/src/pages/intake-form/tabs.tsx
chelsea-EYDS: We can proceed as is for now and then look at this separately

frontend/src/pages/intake-form/tabs.tsx
leon-ey: From Chevonne

image

frontend/src/pages/intake-form/validation.ts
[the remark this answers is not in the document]
leon-ey: From Chevonne:
image

I assume we just apply this answer to both fields or just ignore `indiginousLiving`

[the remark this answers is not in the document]
chelsea-EYDS: Yes I will follow up with Noel re: what to do about the existing data in our db for this one.

frontend/src/pages/intake-form/tabs.tsx
chelsea-EYDS: Sorry - can we still fix the label on line 327?

chelsea-EYDS (approved):
Approving. - TODOs: 
Update functions values
Update travel preferences fields

PR #581, opened 2025-03-18
[TC-1112] Feat/chips frontend

fw-noel:
[TC-1112](https://bcdevex.atlassian.net/browse/TC-1112)

Remaining todos:
- Hiding of banner when necessary

### DIFF
--- a/frontend/src/components/profile/details/ProfileDetails.tsx
+++ b/frontend/src/components/profile/details/ProfileDetails.tsx
@@ -8,6 +8,10 @@ import { DialogUI } from '@/components/ui';
 import { ProfileEditForm } from '../forms';
 import usePersonnel from '@/hooks/usePersonnel';
 import { useRoleContext } from '@/providers';
+import { Banner } from '@/components/ui/Banner';
+import { BannerType } from '../../../common/enums/banner-enum';
+import { differenceInDays } from 'date-fns';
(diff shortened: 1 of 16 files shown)


frontend/src/components/profile/details/ProfileDetails.tsx
```diff
@@ -52,6 +56,81 @@ export const ProfileDetails = () => {
...
+                      <p className="text-sm text-errorRed">
+                        This member&apos;s profile cannot be found in the government
+                        system.
+                      </p>
+                    ) : (
+                      <p className="text-sm text-errorRed">
+                        This member has some profile information flagged for review
+                        as they may contain errors. Please see the red tooltips for
+                        details and make necessary edits.&nbsp; The details with an
+                        exclamation icon are directly sourced from this member&apos;s
```
chelsea-EYDS, line 91: The banner seems to contradict functionality since employeeID and paylistID are not editable.

chelsea-EYDS (approved):
Looks good. One comment re: editable fields and banner could use some padding but otherwise approved :)

Screenshot 2025-03-18 at 10 09 40 PM

frontend/src/components/profile/details/ProfileDetails.tsx
fw-noel: I do think those are ones we have to take CHIPS at face value for.

PR #598, opened 2025-03-21
[TC-974] Enable Export of Data From Database

brrankinp:
Implements functionality for exporting of ECMR and BCWS personnel information as CSV-formatted downloadable files.  This process involves the following adjustments:

1. Addition of new methods to EMCR and BCWS services which query their respective TypeORM entities and join detailed personnel, location and recommitment information from their respective database tables.

2. Addition of new methods to EMCR and BCWS controllers which receive these service methods, convert the output to CSV format using the NPM library json-2-csv, and return streamable file versions of these outputs for download.

3. To make column headers more readable, new "enumerations" (technically consts) have been added to the enums folders for EMCR and BCWS which are used in the controller methods to rename the internal DB names for the column headers to human-readable versions.

### DIFF
--- a/backend/src/common/enums/bcws/csv-headers.ts
+++ b/backend/src/common/enums/bcws/csv-headers.ts
@@ -1,0 +1,161 @@
+/*
+NOTE: this is not strictly an enumeration but is kept in the enums/bcws folder
+for convenience, this const is used to assign human-readable headers to CSV output
+using preassigned key-value pairings
+
+Any future adjustments to the CSV output data will require adding the appropriate
+header replacements to this const (with getRawMany(), all field names are generated
+based on their respective DB table/column (or TypeORM alias) name regardless of joins;
+for instance, personnel.first_name becomes 'personnel_first_name')
+
+Columns labelled "Internal Use" are linking PKs and the like, they don't have much
+reporting value but TypeORM's select logic makes excluding them prohibitively difficult
+so they're labelled for convenient removal for future reporting purposes
+*/
+
+export const BcwsCsvHeaders = [
+  {
+    field: 'bcws_personnel_personnel_id',
+    title: 'BCWS Personnel ID (Internal Use)',
+  },
+  { field: 'bcws_personnel_status', title: 'Status' },
+  { field: 'bcws_personnel_date_applied', title: 'Date Applied' },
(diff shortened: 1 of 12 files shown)


backend/src/common/enums/bcws/csv-headers.ts
```diff
@@ -0,0 +1,161 @@
...
+for instance, personnel.first_name becomes 'personnel_first_name')
+
+Columns labelled "Internal Use" are linking PKs and the like, they don't have much
+reporting value but TypeORM's select logic makes excluding them prohibitively difficult
+so they're labelled for convenient removal for future reporting purposes
+*/
+
+export const BcwsCsvHeaders = [
+  {
+    field: 'bcws_personnel_personnel_id',
```
fw-noel, line 18: Don't need this, as personnel id is already in there

backend/src/common/enums/bcws/csv-headers.ts
```diff
@@ -0,0 +1,161 @@
...
+  },
+  {
+    field: 'personnel_work_location',
+    title: 'Work Location ID (Internal Use)',
+  },
+  {
+    field: 'personnel_home_location',
+    title: 'Home Location ID (Internal Use)',
+  },
+  { field: 'home_loc_id', title: 'Home Location ID (Internal Use)' },
```
fw-noel, line 120: I would simplify these by just doing name, region, fire centre, fire zone. No need for id, or the other ids above

fw-noel (dismissed):

fw-noel:
Approved so we can test in the remote environments. However, please remove the @ Public in a future PR, and the `-test` from the endpoint

frontend/src/hooks/useExportToCSV.tsx
```diff
@@ -0,0 +1,20 @@
+import { useAxios } from './useAxios';
+
+export const useExportToCSV = () => {
```
fw-noel, line 3: I'd recommend passing in a program here as a parameter. You can then just have a singular `exportCsv` function that figures out which program it is and calls the appropriate endpoint.

Then, in Dashboard, you can call
```
const { exportCSV } = useExportToCSV(program);
```

And that way, the logic on the button won't require an if/else, you can just let the hook do that work.

PR #622, opened 2025-03-25
[TC-1052] fix: rename Advance Planning Unit

leon-ey:
Description:
Renamed every instance (except for enum names) of `Advanced Planning Unit` to `Advance Planning Unit`

Note:
Might need to re-seed the DB because the insert query in `queries.ts` was changed from `AdvanceD Planning Unit`.


chelsea-EYDS (changes requested):
We're going to need a migration to change this in the database

leon-ey:
> We're going to need a migration to change this in the database

Great catch, I didn't think about it.

Added a migration to change the name:
image

chelsea-EYDS (approved):

PR #626, opened 2025-03-25
[FIX] prevent day picker from submitting

fw-noel:
Objective:
Calendar selections were triggering early submissions


chelsea-EYDS (approved):
nice

PR #627, opened 2025-03-26
[FIX] fix for when restarting recom for member without recom entry

fw-noel:
Objective:
When a member doesn't have a recommitment for the year (inactive last year, or whatever), clicking the restart recom button results in an error. This changes that


backend/src/recommitment/recommitment.service.ts
```diff
@@ -62,56 +64,91 @@ export class RecommitmentService {
...
-    const otherProgram = key === Program.BCWS ? Program.EMCR : Program.BCWS;
-    const personnel = await this.personnelService.findOneById(id);
-    const otherProgramPersonnel = otherProgram === Program.BCWS ?
-      personnel.bcws :
-      personnel.emcr;
-    let otherProgramStatusReset = false;
-    if (otherProgramPersonnel) {
-      const recommitmentOtherProgram = await this.recommitmentRepository.findOne({
-        where: {
+    });
```
chelsea-EYDS, line 81: Is this going to affect recommitment as well as reinitiation?

fw-noel: This should only happen when you click "Restart Recommitment" as a coordinator, which i think only happens during the reinit period

chelsea-EYDS (approved):

PR #639, opened 2025-03-28
[TC-1172] Program validation + Home Location

chelsea-EYDS:
[TC-1172](https://bcdevex.atlassian.net/browse/TC-1172)

Objective:
- Fixes home location + program validation message visibility
- change font color
- button submits single click
- fix splash button


leon-ey (approved):
Works as intended, tested on my end:

image

PR #641, opened 2025-03-28
[TC-1137] UI fixes

chelsea-EYDS:
[TC-1137](https://bcdevex.atlassian.net/browse/TC-1137)

Objective:


leon-ey (commented):
The only thing I noticed is that the `Delete` button is a bit misaligned with languages/skills picker, but otherwise everything else looks good!

image

chelsea-EYDS:
Fixed the commented issue. Merging

PR #642, opened 2025-03-28
[TC-1202] submit button not registering first click

chelsea-EYDS:
[TC-1202](https://bcdevex.atlassian.net/browse/TC-1202)

Objective:
- Bugfix for button not registering first click


leon-ey (approved):
Tested, works great!

image

PR #647, opened 2025-03-29
[CHORE] reverting download button remove

chelsea-EYDS:
Objective:
- adding the download button back to main


chelsea-EYDS:
Merging - previously approved

PR #649, opened 2025-03-31
[FIX] export query

fw-noel:
Objective:
This is a working download on Dev with 3500 members, resulting in a 2.6mb CSV file.

Further todos:
- Ensure Recommitment details filled out, only should be program specific (left join include clause on program)
- Clean up headers if needed
- Write BCWS equivalent endpoint; remove EMCR specific fields
- No @Public
- Ensure frontend still works fine


chelsea-EYDS (approved):
Approving but need @fw-noel approval prior to merge

chelsea-EYDS (approved):

PR #651, opened 2025-03-31
[TC-1103] update ministry abbreviations and names

fw-noel:
[TC-1103](https://bcdevex.atlassian.net/browse/TC-1103)

Objective:
In the backend, we still have the old abbreviations so as to not change existing data, but we introduce the new acronyms to allow their use.
In the frontend, we replace so that old abbreviations can no longer be selected
We also set the names properly.


backend/src/common/enums/ministry.enum.ts
```diff
@@ -53,28 +58,34 @@ export enum MinistryName {
...
   TACS = 'Tourism, Arts, Culture and Sport',
-  MOTI = 'Transportation and Infrastructure',
+  MOTI = 'Transportation and Transit',
   WLRS = 'Water, Land and Resource Stewardship',
-  OTHER = 'Other'
+  OTHER = 'Other',
+  ECS = 'Energy and Climate Solutions',
+  MCM = 'Mining and Critical Minirals',
+  INF = 'Infrastructure',
+  MCFD = MCF,
```
chelsea-EYDS, line 67: Should we remove MCF since we're renaming the enum in the db?

backend/src/database/migrations/1743013336475-rename-certs-migration.ts
chelsea-EYDS, line 1: This has already ran on prod - will renaming cause issues?

chelsea-EYDS (approved):

PR #653, opened 2025-03-31
[TC-1200] adjustment to availability query

chelsea-EYDS:
[TC-1200](https://bcdevex.atlassian.net/browse/TC-1200)

Objective:

Currently if you are searching for a member by an available date range, the member will only show up if the confirmed until date is after the the [query.to](http://query.to/) date. We allow partial matches so this should be changed to include users whose confirmed until date is greater than or equal to the query.from date.

### DIFF
--- a/frontend/src/components/table/classes.ts
+++ b/frontend/src/components/table/classes.ts
@@ -9,6 +9,8 @@ export const tableClass = {
 
 export const getUnionMembershipClass = (value?: string) => {
   switch (value) {
+    case UnionMembership.OTHER:
+      return `${tableClass.classificationClass} text-warning bg-warningBannerLight border-warningDark`;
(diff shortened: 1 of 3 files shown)


frontend/src/components/table/classes.ts
```diff
@@ -9,6 +9,8 @@ export const tableClass = {
 
 export const getUnionMembershipClass = (value?: string) => {
   switch (value) {
+    case UnionMembership.OTHER:
+      return `${tableClass.classificationClass} text-warning bg-warningBannerLight border-warningDark`;
```
chelsea-EYDS, line 13: Seems like this changes was un-done so I redid :)

fw-noel (approved):

